PR #218, opened 2026-06-16
fix(egc-memory): load observations from state store

NITESH-DTU:
## What Changed

This PR addresses an issue where `compress_observations` could return `No raw observations found to compress.` even when hook events were actively being generated.

During investigation, I found that:

* `compress_observations` reads observations via `loadRawObservations()`, which currently only consults `observations.jsonl`.
* The active hook pipeline (`observe-runner.js` → `state-db-writer.js`) persists runtime events into the SQLite state store (`state.db`).

As a result, fresh installs may accumulate events in SQLite without ever populating `observations.jsonl`, causing compression to find no observations.

This PR:

* Adds SQLite-backed observation loading from the `events` table.
* Updates `loadRawObservations()` to attempt SQLite loading first.
* Falls back to the existing JSONL implementation when no SQLite observations are found.

This preserves backward compatibility while restoring expected behavior on fresh installs.

## Why This Change

Producer path:

```text
observe-runner.js
    ↓
teeEventToStateDb()
    ↓
state-db-writer.js
    ↓
insertRuntimeEvent()
    ↓
SQLite events table
```

Consumer path before this change:

```text
compress_observations
    ↓
loadRawObservations()
    ↓
observations.jsonl only
```

This PR attempts to reconnect those paths.

## Testing Done

* [x] Traced the observation producer and consumer paths to identify the source mismatch.
* [x] Verified that the active hook pipeline writes runtime events to the SQLite state store.
* [x] Confirmed that `compress_observations` only consulted `observations.jsonl` before this change.
* [x] Implemented SQLite-backed observation loading with JSONL fallback.
* [x] Verified that the modified TypeScript code compiled successfully before reaching the existing Windows-specific `chmod` build issue.
* [x] Preserved the legacy JSONL path to maintain backward compatibility.
* [ ] Full automated test suite execution on Windows (blocked by the existing `chmod` build script issue unrelated to this change).

## Notes

I intentionally kept this change minimal and backward-compatible.

Potential follow-ups may include:

* Applying `since` filtering to the SQLite query.
* Aligning SQLite observation filtering with the existing JSONL filtering logic.
* Clarifying whether JSONL should remain a supported observation source going forward.

I am happy to make follow-up changes if maintainers would prefer a different approach.

---
## Summary by cubic
Fixes missing observations during compression by reading from the SQLite state store first, with a fallback to `observations.jsonl`. Applies the `since` filter, respects `limit`, and aligns the state DB path with `index.ts` for fresh installs while keeping backward compatibility.

- **Bug Fixes**
  - `loadRawObservations()` now queries the `events` table in `state.db` (resolved via `STATE_STORE_RELATIVE_PATH` or `EGC_STATE_DB`) before `observations.jsonl`.
  - SQL applies the `since` filter and respects `limit`; results are ordered by timestamp desc.
  - Uses `sqlite3` + `sqlite` to load events and maps payloads into `RawObservation` fields.

<sup>Written for commit fd3a27d703fd66171b6d4189e1ea9d0d671a8f01. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mcp/servers/egc-memory/src/compress.ts
+++ b/mcp/servers/egc-memory/src/compress.ts
@@ -6,6 +6,8 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { z } from "zod";
+import sqlite3 from "sqlite3";
+import { open } from "sqlite";
 
 // ─── Types ────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


Fmarzochi:
Great fix, @NITESH-DTU! I made a few small adjustments to your branch so CI stays green:

1. **`STATE_STORE_RELATIVE_PATH` constant** — replaced the hardcoded `".gemini/egc/state.db"` string with a named constant that mirrors the one already in `index.ts`, keeping both files in sync.

2. **`since` filter** — the parameter was received but not used in the SQL query. Added a `WHERE (? IS NULL OR timestamp >= ?)` clause so timestamp filtering actually works.

3. **Indentation** — fixed the indentation of the new block inside `loadRawObservations`.

The logic and approach are yours — these were purely mechanical fixes. Thanks for tracking down the root cause!

Fmarzochi:
Hey @NITESH-DTU, thanks again for the fix!

A few things worth documenting for your next contribution:

**What I changed after your PR was submitted**

Your core logic was correct. I made three mechanical adjustments to get CI green:

1. Extracted a `STATE_STORE_RELATIVE_PATH` constant instead of the inline string `".gemini/egc/state.db"` — mirrors the constant already in `index.ts` so both files stay in sync if the path ever changes.

2. Added the `since` timestamp filter to the SQL query. You received the parameter but it was not being passed to `WHERE`, so timestamp filtering had no effect.

3. Fixed indentation inside `loadRawObservations`.

The idea and the root cause diagnosis are entirely yours.

**DCO sign-off for next time**

This project enforces the [Developer Certificate of Origin](https://developercertificate.org/). Every commit needs a `Signed-off-by` line. The easiest way is:

```bash
git commit -s -m "your message here"
```

The `-s` flag appends it automatically. If you forget on a commit that has not been pushed yet:

```bash
git commit --amend -s
```

If you already pushed, create a new commit with the sign-off rather than amending.

**One more thing**

Make sure the email in your local git config matches a verified email on your GitHub account:

```bash
git config user.email
```

That email is what GitHub uses to credit you as a contributor. If it does not match, your name may not appear in the contributors list.

Looking forward to your next PR!

PR #227, opened 2026-06-16
fix(egc-memory): prevent recompression of SQLite observations

NITESH-DTU:
This follow-up addresses the remaining recompression issue for observations loaded from the SQLite state store.

Changes:
- Added state-store migration `005_events_compressed_at` to track compression status on events.
- Updated SQLite observation loading to skip events where `compressed_at` is already set.
- Updated `replaceObservation()` to mark SQLite-backed observations as compressed after successful compression while preserving the existing JSONL fallback behavior.

With this change, observations originating from SQLite are no longer reprocessed on subsequent compression runs.

This PR builds upon the work merged in #218.
Closes the remaining recompression issue discussed in #218.

---
## Summary by cubic
Prevents recompression of SQLite-backed observations by scoping reads to the current repo, filtering invalid payloads, and tracking processed events via `events.compressed_at`. Keeps the JSONL fallback for projects without the state DB.

- **Bug Fixes**
  - Read from `.gemini/egc/state.db` when present; scope to the repo via `sessions.repo_root` (git root); return only `compressed_at IS NULL`; apply `since` in SQL; skip malformed JSON; fall back to JSONL.
  - When compressing, set `events.compressed_at`; run an idempotent `ALTER TABLE ... ADD COLUMN` in read/write for pre-005 DBs; if no row updated, write to JSONL.
  - Resolve the DB via `STATE_STORE_RELATIVE_PATH`; support `EGC_STATE_DB`.
  - Tests updated to assert migration 005.

- **Migration**
  - Add `compressed_at` to `events` and create index `idx_events_compressed_at_timestamp` on `(compressed_at, timestamp DESC)` in `005_events_compressed_at`.

<sup>Written for commit 143a4a760df28bbc31df5fd1093101cc6dc06110. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Observation compression now prioritizes an SQLite-backed state database, loading uncompressed event records with optional time filtering and result limits before falling back to the existing line-based storage.
  * Updates now mark events with a `compressed_at` timestamp to reflect processing status.

* **Chores**
  * Added a schema migration to introduce the `compressed_at` column on events and an index to improve timestamp-based lookups.
  * Compression updates now write to SQLite when available, with a safe fallback to the prior file-based replacement flow.

### DIFF
--- a/mcp/servers/egc-memory/src/compress.ts
+++ b/mcp/servers/egc-memory/src/compress.ts
@@ -123,7 +123,17 @@ function resolveStateStoreDbPath(): string {
   return path.join(homeDir, STATE_STORE_RELATIVE_PATH);
 }
 
+function resolveProjectRoot(projectPath: string): string {
+  let dir = projectPath;
+  while (dir !== path.dirname(dir)) {
+    if (fs.existsSync(path.join(dir, ".git"))) return dir;
+    dir = path.dirname(dir);
+  }
+  return projectPath;
+}
+
 async function loadRawObservationsFromStateDb(
+  projectPath: string,
   limit: number = 50,
   since?: string
 ): Promise<RawObservation[]> {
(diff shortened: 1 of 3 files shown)


Fmarzochi:
Hi @NITESH-DTU, thank you for this follow-up fix!

Before this can be merged, there are two things to address:

**1. DCO sign-off missing on your latest commit**

Commit `efa1451d` does not have a `Signed-off-by` line. The DCO check will fail when CI runs.

Please amend it:
```bash
git commit --amend -s
git push --force-with-lease
```

The line must match exactly:
```
Signed-off-by: Your Full Name <your@email.com>
```

**2. Tests required for the new behavior**

The project requires tests for all new behavior. Your PR adds:
- Filtering of already-compressed observations (`compressed_at IS NULL` in the query)
- Marking observations as compressed in SQLite after `replaceObservation`

Please add test coverage for both paths in `tests/` before this can be merged. See `tests/compress.test.js` for the existing pattern.

Once both are addressed, a maintainer will approve the CI run and review the code.

Fmarzochi:
Hi @NITESH-DTU, adding some context to the previous comment.

**Why the DCO sign-off is required**

EGC enforces the [Developer Certificate of Origin](https://developercertificate.org/) on every commit. This is a project governance rule, not a GitHub setting: it means that by signing off, you certify that you wrote the code and have the right to contribute it under the MIT license. The `dco.yml` workflow checks every commit in every PR and will block the merge if any commit is missing the signature.

This rule applies to all contributors without exception, including maintainers. Going forward, every commit you open in a PR to EGC must include this line.

**How to sign commits going forward**

Always use the `-s` flag when committing:

```bash
git commit -s -m "fix(egc-memory): prevent recompression of SQLite observations"
```

This automatically appends the required line to your commit message:

```
Signed-off-by: Nitesh Kumar <Niteshkaushik200@gmail.com>
```

**How to fix the unsigned commit in this PR**

Your latest commit (`efa1451d`) is missing the sign-off. To fix it:

```bash
git commit --amend -s
git push --force-with-lease origin <your-branch-name>
```

After the push, the DCO check will pass on the next CI run.

**About the tests**

The project runs `node tests/run-all.js` on every PR across Linux, macOS, and Windows with Node 20 and 22. New behavior needs test coverage before merging. See `tests/compress.test.js` for the existing pattern: it imports from the compiled build and uses `assert` from Node's standard library. No external test framework needed.

Looking forward to the fixes!

Fmarzochi:
Hi @NITESH-DTU, thanks again for this fix. Given the 72-hour review window, I went ahead and pushed directly to your branch (maintainerCanModify is enabled) so this could land without further delay. The core idea you implemented was solid.

Here is what was adjusted on top of your work:

**1. Project scope filter**
The original query read from the global state DB without filtering by project, so a compression run on project A could process events from project B. The query now does an INNER JOIN on `sessions.repo_root` and filters by the resolved project root of the caller.

**2. Malformed JSON rows**
When `JSON.parse` failed, the code fell back to `payload = {}` and returned an empty observation, which then got marked as compressed in the DB. Those rows are now filtered out before mapping, so only rows with valid observation content are returned.

**3. Migration guard**
The code accessed `compressed_at` without ensuring migration 005 had been applied, which would crash on any existing state DB. An idempotent `ALTER TABLE ADD COLUMN` now runs before each query, so the code works safely against older DBs.

**4. Composite index on migration 005**
The migration added the column but no index. The `WHERE compressed_at IS NULL ORDER BY timestamp DESC` query would do a full table scan as the events table grows. Migration 005 now also creates `idx_events_compressed_at_timestamp`.

Everything is in one clean squash commit with DCO sign-off. The PR is ready to merge.

If you spot other edge cases or bugs anywhere in the codebase, please open an issue or a new PR. Your contributions are very welcome here.

Issue #386, opened 2026-06-22
bug: AttributeError when Claude response starts with a tool_use block


Fmarzochi:
## Description

In `src/llm/providers/claude.py`, line 90 unconditionally reads `response.content[0].text`. When Claude's response starts with a `tool_use` block (no preceding text), that block has no `.text` attribute, raising an `AttributeError`.

## Affected file

`src/llm/providers/claude.py`, lines 80-90

```python
if response.content[0].type == "tool_use":
    tool_calls = [...]         # correctly extracts tool call

return LLMOutput(
    content=response.content[0].text if response.content else "",  # crashes here
    ...
)
```

## How to reproduce

Call any EGC tool-enabled prompt that makes Claude respond with only a tool call and no leading text. The `content[0]` block will be `tool_use`, which has no `.text` attribute.

## Expected behavior

`content` should be the text from the first `text`-type block, or an empty string if no text block exists.

## Suggested fix

```python
content = next(
    (block.text for block in response.content if block.type == "text"),
    "",
)
```

Replace the `response.content[0].text` expression with the snippet above.

krishna3554:
Hi, I can help you with the issue. Please assign me the issue.

Fmarzochi:
Hey @krishna3554, welcome to EGC!

The issue is yours. The fix is straightforward — replace the `response.content[0].text` expression in `src/llm/providers/claude.py` with the suggested snippet in the description:

```python
content = next(
    (block.text for block in response.content if block.type == "text"),
    "",
)
```

A few things to keep in mind before opening a PR:

- Sign your commit with `-s` (DCO required): `git commit -s`
- Run the test suite locally before pushing
- Feel free to ask questions here if anything is unclear

Looking forward to your PR!

Fmarzochi:
Hey @krishna3554 — just a heads up: if you'd like your contributions to count officially in the EGC repository, you'll need to sign our CLA.

Read [.github/CLA.md](https://github.com/Fmarzochi/EGC/blob/main/.github/CLA.md) and post the required comment on any open PR to complete the signing process.

Your co-authorship on commit `caeb4bd1` is already in the git history — signing the CLA just makes it fully official for future contributions. We'd love to see you take on one of the open issues: **#391**, **#392**, or **#393**! 🙌

PR #387, opened 2026-06-22
Fix Claude response bug

Ap-0007:
## What was broken
AttributeError when Claude response starts with a tool_use block
## What changed
Added check for tool_use block before accessing text attribute
## How to test
Test with a Claude response starting with a tool_use block

---

---
## Summary by cubic
Fixes a crash (AttributeError) when Claude responses start with a tool_use block by checking for a text attribute before reading it. If no text exists, we return an empty content string while preserving tool_calls, model, usage, and stop_reason.

<sup>Written for commit 0017102db7eb23bb148ad3fa582b20e89d253893. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/387?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of Claude provider by adding defensive checks when processing API responses, ensuring the system gracefully handles edge cases where response content may be missing or malformed.

### DIFF
--- a/src/llm/providers/claude.py
+++ b/src/llm/providers/claude.py
@@ -86,8 +86,12 @@ class ClaudeProvider(LLMProvider):
                         )
                     ]
 
+                        if response.content and hasattr(response.content[0], 'text'):
+                text = response.content[0].text
+            else:
+                text = ''
             return LLMOutput(
-                content=response.content[0].text if response.content else "",
+                content=text,
                 tool_calls=tool_calls,
                 model=response.model,
                 usage={


Fmarzochi:
Hey @krishna3554, welcome and thanks for tackling this!

The approach is on the right track, but there are a couple of things to fix before this is ready to merge.

**Issue 1 - Indentation is broken**

Look at this block:

```python
                        if response.content and hasattr(response.content[0], 'text'):
                text = response.content[0].text
            else:
                text = ''
```

The `if` line has extra indentation — it looks like it ended up inside the `for` loop above. Python will raise an `IndentationError` here. Make sure all three lines (`if`, `text = ...`, `else`, `text = ''`) are at the same level, outside any loop.

**Issue 2 - The logic still has a bug**

Even after fixing the indentation, using `hasattr(response.content[0], 'text')` only checks the *first* block. If Claude responds with a `tool_use` block first and a `text` block second, you will still return an empty string — missing the actual text content.

Think about it this way: you want to find the first block where `block.type == "text"`, regardless of its position. Python has a built-in way to do exactly that with `next()` and a generator expression.

Try replacing your `if/else` with something like:

```python
content = next(???, "")
```

Fill in the `???` with a generator that iterates over `response.content` and filters by `block.type == "text"`. Then use `content` in the `LLMOutput` call.

Give it a shot and push the updated commit — happy to review again!

Fmarzochi:
Hey @Ap-0007, first of all — thank you so much for jumping on this so quickly! It means a lot that you took the initiative to open a fix this fast.

I owe you an apology though: we had already assigned issue #386 to @krishna3554 before your PR came in, and we want to give them the chance to work through it. So we'll be closing this PR and letting them carry it to the finish line.

That said, we really don't want to lose you — your speed and willingness to contribute are exactly what this project needs. We just opened issue #388, which has a very similar bug in the OpenAI provider: a `json.JSONDecodeError` that isn't handled properly in `src/llm/providers/openai.py`. It's small, focused, and a great fit.

Would you be up for taking it on? No pressure at all — but if you are, we'll be right here to review and guide you through it. 🙏

Fmarzochi:
Almost there, @krishna3554 — but one more thing before this can merge.

**The CI is failing: `codecov/patch` — 50% coverage (target: 80%)**

You fixed the code, but the diff has no tests covering the new path. Codecov is measuring coverage only on the lines you changed, and half of them are untested.

You need to add a test that exercises the bug you fixed — a Claude response where `content[0]` is a `tool_use` block with no preceding `text` block.

There are no existing tests for `src/llm/providers/claude.py`, so you'll need to create one. Here's the structure to follow:

```python
# tests/test_claude_provider.py
from unittest.mock import MagicMock, patch
from llm.providers.claude import ClaudeProvider
from llm.core.types import LLMInput, Message, MessageRole

def test_generate_returns_empty_content_when_response_starts_with_tool_use():
    """content should be '' when first block is tool_use, not AttributeError."""
    mock_response = MagicMock()
    mock_response.stop_reason = "tool_use"

    tool_block = MagicMock()
    tool_block.type = "tool_use"
    tool_block.name = "some_tool"
    tool_block.id = "call_123"
    tool_block.input = {"key": "value"}

    mock_response.content = [tool_block]
    mock_response.model = "claude-3-5-sonnet-20241022"
    mock_response.usage.input_tokens = 10
    mock_response.usage.output_tokens = 5

    # fill in the rest based on how ClaudeProvider is initialized
    ...
```

The exact setup depends on how `ClaudeProvider.__init__` works — look at that file and adapt. The key assertion is:

```python
assert result.content == ""
assert result.tool_calls is not None
```

Add the test, run it locally (`pytest tests/test_claude_provider.py`), and push. That should bring coverage up above 80%.

Fmarzochi:
Hey @Ap-0007, thank you again for your speed and enthusiasm — it's genuinely appreciated! 🙏

We ended up going with a separate fix for this one since the issue had already been assigned to @krishna3554 before your PR came in, and their approach needed a slightly different implementation to be fully correct.

But we absolutely want you on the next challenge! Issue #388 is waiting for you — it's a very similar bug in the OpenAI provider, same type of missing error handling, and just as approachable. We're ready to review and guide you through it.

Hope to see your PR soon!

Issue #393, opened 2026-06-22
bug: json.JSONDecodeError unhandled in OpenRouterProvider tool_calls argument parsing


Fmarzochi:
## Description

`OpenRouterProvider` inherits `generate()` from `OpenAIProvider`, which contains this line at `src/llm/providers/openai.py:92`:

```python
arguments={} if not tc.function.arguments else json.loads(tc.function.arguments),
```

If the OpenRouter API returns malformed JSON in `function.arguments` (common with cheaper models), `json.loads` raises `json.JSONDecodeError`. This is not caught and propagates as a raw exception instead of a structured `LLMError`, breaking the provider contract for all OpenRouter calls.

This is the same pattern as the bug in issue #388 (OpenAI) — but because OpenRouter inherits the method unchanged, it is affected too.

## Expected behavior

Malformed arguments should fall back to `{}` gracefully.

## Suggested fix

```python
try:
    args = json.loads(tc.function.arguments) if tc.function.arguments else {}
except json.JSONDecodeError:
    args = {}
```

Then use `args` in the `ToolCall` constructor.

## Affected file

`src/llm/providers/openai.py`, line 92 (also affects `OpenRouterProvider` via inheritance)

Issue #392, opened 2026-06-22
bug: TypeError when msg.content is None in GeminiProvider message building


Fmarzochi:
## Description

In `src/llm/providers/gemini.py`, line 171, `types.Part.from_text()` is called with `msg.content` directly:

```python
else:
    parts.append(types.Part.from_text(text=msg.content))
```

When an assistant message has `content=None` (valid at runtime — the `str` type hint is not enforced), the Gemini SDK raises:

```
TypeError: argument of type 'NoneType' is not iterable
```

## How to reproduce

Pass an assistant `Message` with `content=None` and no `tool_calls` to `GeminiProvider.generate()`.

## Expected behavior

`None` content should fall back to an empty string.

## Suggested fix

```python
else:
    parts.append(types.Part.from_text(text=msg.content or ""))
```

## Affected file

`src/llm/providers/gemini.py`, line 171

binggao1230:
I can take this. I'll keep it scoped to treating None message content as an empty string in GeminiProvider and adding a focused regression test.

Fmarzochi:
Assigned to you @gaoflow! Appreciate you picking this one up too. Same process applies.

Fmarzochi:
Fixed in PR #405. Thanks @gaoflow!

Issue #391, opened 2026-06-22
bug: TypeError when temperature=None in OllamaProvider


Fmarzochi:
## Description

In `src/llm/providers/ollama.py`, line 67, the code does arithmetic on `input.temperature` without checking for `None` first:

```python
if abs(input.temperature - 1.0) > 1e-9:
```

If a caller passes `temperature=None`, Python raises:

```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

## How to reproduce

```python
provider.generate(LLMInput(messages=[...], model="llama3", temperature=None))
```

## Expected behavior

When `temperature` is `None`, the block should be skipped gracefully.

## Suggested fix

```python
if input.temperature is not None and abs(input.temperature - 1.0) > 1e-9:
    payload["options"] = {"temperature": input.temperature}
```

## Affected file

`src/llm/providers/ollama.py`, line 67

binggao1230:
I can take this. I'll keep it scoped to skipping the Ollama temperature options block when temperature is None and adding a focused regression test.

Fmarzochi:
Assigned to you @gaoflow! Great to see you back. Same process as before -- fork, `git commit -s`, and the CLA is already signed. Feel free to open the PR whenever you are ready.

Fmarzochi:
Fixed in PR #406. Thanks @gaoflow!

PR #394, opened 2026-06-22
Fix content extraction in Claude tool_use response

krishna3554:
## Summary

Fix an `AttributeError` in the Claude provider when a response starts with a `tool_use` block.

The previous implementation assumed that `response.content[0]` always contained a text block and accessed its `text` attribute directly. However, Claude can return a `tool_use` block as the first (or only) content item, which does not have a `text` attribute.

This change updates the content extraction logic to safely retrieve the first available text block from the response and fall back to an empty string when no text block is present.

Fixes #<issue-number>

## Changes

* Replace direct access to `response.content[0].text`
* Extract the first text block from `response.content`
* Return an empty string when the response contains only tool calls

## Testing

* Ran the project test suite successfully
* Verified that text responses continue to work as expected
* Verified that responses beginning with `tool_use` blocks no longer raise an `AttributeError`

---
## Summary by cubic
Fixes an AttributeError in the Claude provider when a response starts with a `tool_use` block by safely extracting the first `text` block from `response.content`, or returning an empty string if none exists. Prevents crashes on tool-only responses and keeps normal text responses unchanged.

<sup>Written for commit b98d6b1befe64f9b5519026b30b11c1874e8cd7c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/llm/providers/claude.py
+++ b/src/llm/providers/claude.py
@@ -87,7 +87,9 @@ class ClaudeProvider(LLMProvider):
                     ]
 
             return LLMOutput(
-                content=response.content[0].text if response.content else "",
+                content=next(
+                    (block.text for block in response.content if block.type == "text"), "",
+                ),
                 tool_calls=tool_calls,
                 model=response.model,
                 usage={


Fmarzochi:
Hey @krishna3554 — this is actually a great fix and your logic is spot on! 🎉

Here's what happened: while you were working on this, we went ahead and implemented the same fix on our end to keep things moving. Your solution is correct — using `next()` with a generator filtered by `block.type == "text"` is exactly the right approach.

The good news: **you're already credited as co-author on the commit that's now in main** ([caeb4bd1](https://github.com/Fmarzochi/EGC/commit/caeb4bd1)). Your name is in the git history and visible on GitHub — that contribution is yours.

We'd love to keep the momentum going with you. There are 3 fresh `good first issue` bugs waiting, all very similar to this one:

- **#391** — `ollama.py`: `TypeError` when `temperature=None`
- **#392** — `gemini.py`: `TypeError` when `msg.content` is `None`
- **#393** — `openrouter.py`: unhandled `JSONDecodeError` in tool call arguments

Pick whichever one catches your eye — we'll be here to review. Thanks again for showing up and delivering a correct fix. That counts for a lot. 🙌

PR #430, opened 2026-06-23
fix: handle null text in ClaudeProvider content extraction

vasu-sachdeva:
﻿## Description

When the Anthropic SDK returns a text block with 	ext=None in edge cases (empty streaming responses, partial responses), content was set to None instead of an empty string. This breaks downstream code that expects a string.

## Fix

Changed lock.text to lock.text or "" on line 94 of src/llm/providers/claude.py, following the same defensive pattern already used on lines 85-86 (getattr(block, "id", "")).

## Testing

All 5 existing tests pass.

Closes #428

---
## Summary by cubic
Fixes content extraction in `ClaudeProvider` by treating null text blocks from the Anthropic SDK as empty strings. Prevents crashes when downstream code expects a string.

- **Bug Fixes**
  - Use `(block.text or "")` when selecting the first text block (covers empty streaming/partial responses).

<sup>Written for commit 6698c33e4a0932419246b94f53b1bc872355a707. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing text content in API responses, ensuring empty strings are returned consistently instead of null values.

PR #431, opened 2026-06-23
fix: handle null message in OllamaProvider content extraction

### LINKED ISSUE #427: bug: OllamaProvider crashes with AttributeError when API returns message: null
## Description

`src/llm/providers/ollama.py` line 79 uses `.get("message", {})` to safely access the message field, but this default only applies when the key is **absent**. If Ollama returns `{"message": null}`, `.get("message", {})` returns `None` (not `{}`), and the subsequent `.get("tool_calls")` call raises `AttributeError`.

## Affected code

```python
# src/llm/providers/ollama.py, line 79
if result.get("message", {}).get("tool_calls"):  # crashes if message is null
```

## Reproduction

```python
result = {"message": None}
result.get("message", {}).get("tool_calls")  # AttributeError: 'NoneType' has no 'get'
```

## Fix hint

Replace:
```python
result.get("message", {}).get("tool_calls")
```
With:
```python
(result.get("message") or {}).get("tool_calls")
```

The same pattern should be applied consistently to the `content` extraction on line 76.

## Steps

1. Fork the repo
2. Fix `src/llm/providers/ollama.py` lines 76 and 79
3. Run `python -m pytest tests/test_ollama_provider.py`
4. Open a PR referencing this issue

vasu-sachdeva:
﻿## Description

When the Ollama API returns {"message": null}, 
esult.get("message", {}) returns None instead of {} because the key *exists* but is null. This causes AttributeError: 'NoneType' has no attribute 'get' on both:

1. Line 76 - content extraction: .get("content", "")
2. Line 79 - tool_calls check: .get("tool_calls")

## Fix

Changed both lines from 
esult.get("message", {}) to (result.get("message") or {}), which safely falls back to an empty dict even when the value is None.

## Testing

Existing test passes:
`
tests/test_ollama_provider.py::TestOllamaProviderPayload::test_none_temperature_omits_options PASSED
`

Closes #427

---
## Summary by cubic
Fixes a crash in `OllamaProvider.generate` when the API returns `{"message": null}` by defaulting `message` to `{}`. Content and `tool_calls` extraction now safely handle a null `message` without raising `AttributeError`.

<sup>Written for commit 30d34d94a918f10e1f40cebd56f32661fbc8c275. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of LLM provider response handling to gracefully manage edge cases in result data processing.

### DIFF
--- a/src/llm/providers/ollama.py
+++ b/src/llm/providers/ollama.py
@@ -73,10 +73,10 @@ class OllamaProvider(LLMProvider):
             with urllib.request.urlopen(req, timeout=60) as response:
                 result = json.loads(response.read().decode("utf-8"))
 
-            content = result.get("message", {}).get("content", "")
+            content = (result.get("message") or {}).get("content", "")
 
             tool_calls = None
-            if result.get("message", {}).get("tool_calls"):
+            if (result.get("message") or {}).get("tool_calls"):
                 tool_calls = [
                     ToolCall(
                         id=tc.get("id", ""),


vasu-sachdeva:

Fmarzochi:
Great catch on this one, @vasu-sachdeva! The `(result.get("message") or {})` pattern is the correct fix -- the default-argument form only triggers when the key is absent, not when its value is `None`. You identified the subtle distinction exactly right.

Same two steps needed before we can merge:

**1. DCO (required for every commit)**
```bash
git commit --amend --signoff
git push --force-with-lease
```

**2. CLA (one-time -- covers both this PR and #430)**
If you already posted it on PR #430, you're done here. Otherwise post:

> I have read the CLA Document and I hereby sign the CLA.

Once both checks are green, this merges.

Fmarzochi:
One more thing: SonarCloud is flagging a **cognitive complexity** issue on `generate()` in `ollama.py`. The function was already at the limit of 15, and using `(result.get("message") or {})` twice pushed it to 16.

The fix is simple -- extract `message` into a variable once, then use it directly on both lines:

```python
message = result.get("message") or {}
content = message.get("content", "")

tool_calls = None
if message.get("tool_calls"):
    tool_calls = [
        ToolCall(
            id=tc.get("id", ""),
            name=tc.get("function", {}).get("name", ""),
            arguments=tc.get("function", {}).get("arguments", {}),
        )
        for tc in message["tool_calls"]
    ]
```

This is better in every way: the null-message guard runs once, the code is less repetitive, and complexity drops back to 15. Note `message["tool_calls"]` is safe here because the `if message.get("tool_calls")` guard already confirmed the key exists.

So the full checklist before merge:
- [ ] `git commit --amend --signoff` and `git push --force-with-lease` (DCO)
- [ ] Post CLA comment (if not done on PR #430 already)
- [ ] Apply the refactor above (SonarCloud)

Fmarzochi:
Applied in edf62d2 on main with your fix (plus the message variable extraction to fix SonarCloud complexity) and Co-authored-by credit. Great catch on the null message vs absent key distinction, @vasu-sachdeva. The DCO signoff was the only blocker, so we applied it directly. You're now a contributor to EGC!

Issue #509, opened 2026-06-29
feat(security): rate-limit MCP tool calls to prevent flooding


Fmarzochi:
## Hardening
The MCP tools have no rate limiting, so a misbehaving or hostile client can flood the server with calls (local DoS against the EGC servers).

## Proposal
Add simple per-tool rate limiting (token bucket or fixed window) in the MCP request path, with a clear error when the limit is exceeded. Make the limit configurable.

## Where
`mcp/servers/egc-guardian/` and `mcp/servers/egc-memory/` - tool dispatch.

Good first issue: contained, testable.

PR #517, opened 2026-06-29
Rate limit MCP tool calls

Ap-0007:
## What was broken
No rate limiting on MCP tools
## What changed
Added rate limiting to prevent flooding
## How to test
Send multiple requests to MCP tools and verify rate limiting

---

---
## Summary by cubic
Adds basic rate limiting for MCP tool calls to prevent flooding and abuse. Caps requests to 100 per 15 minutes.

- **New Features**
  - Integrated `express-rate-limit` in the EGC Guardian chunk router.
  - Enforces a 15-minute window with a max of 100 tool invocations.

<sup>Written for commit 18bc7c8bf68e78020d0352a8305d577b3b64ea2b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/517?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Fmarzochi:
Thank you for picking up issue #509 -- rate limiting tool calls is a legitimate security concern and we are glad someone is thinking about it.

However, this implementation has a few fundamental issues that need to be addressed before it can merge:

**1. `express-rate-limit` is an HTTP middleware -- it does not apply here**

`classifyChunk` is a pure text classification function called internally. It is not an HTTP endpoint. `express-rate-limit` works by wrapping Express.js `(req, res, next)` middleware -- calling `limiter(chunk, {}, callback)` as done in this PR has no effect: the rate limiter receives a string where it expects an HTTP request object, so it will always call the callback immediately without limiting anything.

**2. The `return` statements inside the callback do not propagate**

The original function returned the content type directly. Wrapping the body in a callback means those `return` statements now return from the callback, not from `classifyChunk`. The function always falls through to `return 'text'`, which breaks the entire chunk classification logic.

**3. Rate limiting belongs at the MCP tool handler level, not inside a classifier**

The right place to implement call-rate limiting for MCP tools is in `mcp/servers/egc-guardian/src/index.ts`, inside the `callToolHandler`. You could track call counts per tool name using a simple in-memory `Map` with a time-window reset:

```typescript
const callCounts = new Map<string, { count: number; resetAt: number }>();
const RATE_LIMIT = 100;
const WINDOW_MS = 15 * 60 * 1000;

function checkRateLimit(toolName: string): boolean {
  const now = Date.now();
  const entry = callCounts.get(toolName);
  if (!entry || now > entry.resetAt) {
    callCounts.set(toolName, { count: 1, resetAt: now + WINDOW_MS });
    return true;
  }
  if (entry.count >= RATE_LIMIT) return false;
  entry.count++;
  return true;
}
```

Then at the top of each tool case, call `checkRateLimit(toolName)` and return an error if it returns `false`. No external dependency needed.

**4. Please remove `.agent-ping.md`**

This is an internal agent artifact and should not be committed.

Please also include `Signed-off-by` in your commits (`git commit -s`) for DCO compliance.

The intent behind this PR is good -- the implementation just needs to be redirected to the right layer. Feel free to ask if you have questions about the handler structure!

Fmarzochi:
Hi @Ap-0007 -- thanks for opening this PR and identifying these security gaps! The fixes have been implemented correctly in the TypeScript source files with your co-authorship credited in PR #528. This PR will be closed in favor of #528. Appreciate the contribution!

PR #613, opened 2026-07-03
fix: prevent libuv assertion failure on state store init crash

fuentes71:
Fixes a native Node.js crash (Assertion failed in uv_async.c) on Windows when the EGC state store fails to initialize. Ensures the database is correctly closed, preventing WASM memory leaks before terminating the process.


Fmarzochi:
Hi @fuentes71, thank you for opening this PR!

After reviewing the diff, all 316 000+ lines added are in `.agents/skills/` -- these are EGC's own bundled skill library files that ship with the package. They aren't part of a libuv crash fix.

A fix for the described assertion failure (uv_async.c crash on Windows when the state store fails to initialize) would need changes in `mcp-servers/egc-memory/` -- specifically in the SQLite initialization or the async handle lifecycle.

Could you share what you actually changed to reproduce and fix the crash? If you have a real patch, please open a new PR with only the relevant source files. Happy to help guide the fix.

PR #619, opened 2026-07-03
fix: prevent libuv assertion failure on state store init crash

fuentes71:
Fixes a native Node.js crash (Assertion failed in uv_async.c) on Windows when the EGC state store fails to initialize. Ensures the database is correctly closed, preventing WASM memory leaks before terminating the process.

---
## Summary by cubic
Prevents a native Node.js crash on Windows (libuv assertion in uv_async.c) when the state store initialization fails. Ensures the SQLite WASM DB is always closed and the process exits cleanly to avoid memory leaks.

- **Bug Fixes**
  - Wrap state store setup in try/catch; on error, close the DB before rethrowing.
  - Make DB adapter close() safe and idempotent; persist in try/finally and null out the handle.
  - Use process.exitCode = 1 in `bootstrap-state-db` so the event loop can drain before exiting.

<sup>Written for commit 5e9bafa45b45b3c46c5545c208a6930acd0dd49f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state database startup and shutdown handling to clean up more reliably when errors occur.
  * The app now exits more gracefully on bootstrap failures while preserving the error message output.
  * Database closing is now more defensive, reducing the chance of lingering resources after a failure.

PR #629, opened 2026-07-04
fix: resolve process segfault on state db bootstrap failure

fuentes71:
Fixes a critical issue where a database initialization failure would crash the Node.js process with a Segfault (Access Violation).

- Updated scripts/bootstrap-state-db.js to use process.exit(1) instead of setting process.exitCode to prevent WASM teardown issues during event loop exhaustion.
- Wrapped _persist() in a try-catch block inside db-adapter.js close() method to prevent exceptions from overriding each other and crashing the V8 isolate.

---
## Summary by cubic
Prevents a Node.js segfault when state DB bootstrap fails by exiting immediately with code 1 and making DB shutdown ignore cleanup errors so the original error isn’t masked.

- **Bug Fixes**
  - In `scripts/bootstrap-state-db.js`, call `process.exit(1)` instead of setting `process.exitCode` to avoid WASM teardown crashes when the event loop is exhausted.
  - In `scripts/lib/state-store/db-adapter.js` `close()`, ignore `_persist()` errors and wrap `_db.close()` in try/catch so cleanup can’t hide the original error.

<sup>Written for commit 5c0f476d9b182ac4af7dadbf5d44f7730c98727d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/629?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/bootstrap-state-db.js
+++ b/scripts/bootstrap-state-db.js
@@ -30,7 +30,7 @@ if (require.main === module) {
     })
     .catch(err => {
       process.stderr.write(`[bootstrap-state-db] FAILED: ${err.message}\n`);
-      process.exitCode = 1;
+      process.exit(1);
     });
 }
 
--- a/scripts/lib/state-store/db-adapter.js
+++ b/scripts/lib/state-store/db-adapter.js
@@ -121,8 +121,10 @@ class SqlJsDatabase {
     if (!this._db) return;
     try {
       this._persist();
+    } catch (_) {
+      // Ignore persist errors on close to prevent masking original errors
     } finally {
-      this._db.close();
+      try { this._db.close(); } catch (_) { /* ignore close errors */ }
       this._db = null;
     }
   }


Fmarzochi:
Hi @fuentes71 — a few things to address before this can merge:

**1. CLA not signed (blocks all your PRs)**
None of your PRs can merge until you sign the CLA. The CLA bot has posted a link in this PR — please sign it now, it applies to all your contributions at once.

**2. Regression in `scripts/bootstrap-state-db.js`**
You changed `process.exitCode = 1` back to `process.exit(1)` — this reverts the fix from PR #619, which was specifically opened to prevent the libuv assertion failure (segfault) that `process.exit(1)` causes mid-stream. Please revert this change and keep `process.exitCode = 1`.

**3. SonarCloud: empty catch blocks (S2486) in `scripts/lib/state-store/db-adapter.js` lines 124 and 127**
Empty `catch` blocks silently swallow errors. Either log the error to stderr or restructure. The intent (prevent close() errors from masking persist() errors) is right but the pattern is wrong. This is the correct approach:

```diff
  try {
    this._persist();
+ } catch (err) {
+   process.stderr.write(`[db-adapter] persist error on close: ${err.message}\n`);
  } finally {
-   this._db.close();
+   try { this._db.close(); } catch { /* close errors do not matter here */ }
    this._db = null;
  }
```

Once the regression is reverted and the empty catches are fixed, the logic here is sound.

PR #630, opened 2026-07-04
fix: wrap ping() fetch in try/catch to prevent synchronous throw

fuentes71:
## Problem

The \ping()\ function in \	elemetry.js\ is documented as 'Never throws — telemetry must never break the CLI'. However, if \global.fetch\ throws **synchronously** (e.g. when the runtime replaces fetch with a stub that throws immediately), the exception escapes to the caller.

The existing \.catch(() => {})\ only handles **Promise rejections**, not synchronous throws.

## Fix

Wrapped the \etch()\ call in a \	ry/catch\ block to honour the documented safety contract.

\\\js
// Before
fetch(url, { ... }).catch(() => {});

// After
try {
  fetch(url, { ... }).catch(() => {});
} catch (_) {
  // Telemetry must never break the CLI
}
\\\

## Stress Tests Added

Added \	ests/stress/\ with 48 total scenarios across three files:

| File | Scenarios |
|---|---|
| \state-store.stress.js\ | 15 — high-volume inserts, SQL injection, unicode, memory leak probe |
| \	elemetry.stress.js\ | 22 — malformed payloads, FD leaks, directory collision, unicode paths |
| \db-adapter.stress.js\ | 11 — corruption, double-close, rollback integrity, 5000-row transactions |

All 48 pass. This bug was discovered via the stress test suite.

---
## Summary by cubic
Prevents telemetry `ping()` from ever throwing by wrapping `fetch` in try/catch, and adds a hardened stress test suite across telemetry, state-store, and the DB adapter. Also resolves a CodeQL alert by switching test randomness to `crypto`.

- **Bug Fixes**
  - Wrapped `ping()` fetch in try/catch so sync `global.fetch` errors never escape and telemetry can’t break the CLI.
  - Replaced test randomness with `crypto.randomInt` to address CodeQL “Insecure randomness”.

- **New Features**
  - Added stress tests covering malformed consent files, unicode/edge paths, directory collisions, FD leak probes, DB corruption/rollbacks, large transactions, and high-volume ops.
  - Tightened the suite: FK enforcement assertions, unicode/SQL-injection round‑trip checks, nested `db.transaction()` coverage, and stabilized memory‑leak probe via `global.gc()` where available.

<sup>Written for commit d4cf38c54be93f958d448ad00045a166a3984d8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/630?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry error handling so CLI usage is not interrupted by telemetry failures, including broken fetch behavior and other runtime errors.
  * Strengthened state storage reliability under edge cases such as corrupted data, repeated open/close cycles, rollback scenarios, and large-volume writes.

* **Tests**
  * Added stress tests covering telemetry, database adapter, and state store behavior across a wide range of failure and load conditions.

### DIFF
--- a/scripts/lib/telemetry.js
+++ b/scripts/lib/telemetry.js
@@ -117,10 +117,14 @@ function ping(pagePath, title) {
 
   const url = `${GOATCOUNTER_URL}?${params.toString()}`;
 
-  fetch(url, {
-    method: 'GET',
-    headers: { 'User-Agent': userAgent },
-  }).catch(() => {});
+  try {
+    fetch(url, {
+      method: 'GET',
+      headers: { 'User-Agent': userAgent },
+    }).catch(() => {});
+  } catch (_) {
+    // Telemetry must never break the CLI — swallow all errors silently
+  }
 }
 
 module.exports = {
--- a/tests/stress/db-adapter.stress.js
+++ b/tests/stress/db-adapter.stress.js
@@ -1,0 +1,249 @@
+'use strict';
+
+/**
+ * STRESS TEST: scripts/lib/state-store/db-adapter.js
+ * Tests beyond the standard suite:
+ * - Corrupted DB files (should not segfault)
+ * - Zero-byte DB file
+ * - Truncated DB file
+ * - DB file growing under write pressure
+ * - Close called twice (double-free guard)
+ * - Statement freeing under error conditions
+ */
+
+const assert = require('assert');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+let passed = 0;
+let failed = 0;
+const failures = [];
+
+async function test(name, fn) {
+  try {
+    await fn();
+    console.log(`  ✓ ${name}`);
+    passed++;
+  } catch (err) {
+    console.log(`  ✗ ${name}`);
+    console.log(`    Error: ${err.message}`);
+    failures.push({ name, error: err.message });
+    failed++;
+  }
+}
+
+function createTempDir() {
+  return fs.mkdtempSync(path.join(os.tmpdir(), 'egc-dba-stress-'));
+}
+
+function cleanup(dir) {
+  try { fs.rmSync(dir, { recursive: true, force: true }); } catch (_) { /* ignore */ }
+}
+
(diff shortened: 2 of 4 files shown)


Fmarzochi:
Hi @fuentes71 — same CLA blocker applies here (sign it once and it covers all your PRs). Beyond that, SonarCloud flagged 6 issues across 3 files:

**1. `scripts/lib/telemetry.js` — S4822: redundant `try` + S2486: empty `catch`**
`fetch()` returns a Promise, so `.catch(() => {})` already handles all rejections. The surrounding `try/catch` adds nothing (promise errors never reach a synchronous catch) and introduces an empty catch block. Keep the original form:

```diff
-  try {
-    fetch(url, {
-      method: 'GET',
-      headers: { 'User-Agent': userAgent },
-    }).catch(() => {});
-  } catch (_) {
-    // Telemetry must never break the CLI — swallow all errors silently
-  }
+  fetch(url, {
+    method: 'GET',
+    headers: { 'User-Agent': userAgent },
+  }).catch(() => {});
```

If you want to catch a (theoretical) synchronous throw from `fetch` itself:
```js
Promise.resolve().then(() => fetch(url, { method: 'GET', headers: { 'User-Agent': userAgent } })).catch(() => {});
```

**2. `tests/stress/db-adapter.stress.js` lines 65, 85, 162 — S2699: test cases without assertions**
Each `test()` block must include at least one `assert.*` call so the test can actually fail. A test that only runs code without asserting anything passes trivially even when the code is broken.

**3. `tests/stress/telemetry.stress.js` line 157 — S2699: test case without assertions**
Same issue — add an `assert.ok`, `assert.strictEqual`, or similar to verify the expected outcome.

The stress test suite is a great contribution — these are mechanical fixes to get it over the quality gate.

PR #851, opened 2026-07-18
fix(memory): prevent branch state filename collisions

shirokuma-xv:
## Summary

- derive branch-state filenames from a readable prefix plus a SHA-256 digest of the exact Git ref
- keep legacy sanitized branch filenames readable, while new writes use collision-resistant paths
- keep `egc watch` aligned with the same current-first, legacy-fallback resolution
- add regression coverage for colliding branch names and legacy state migration

## Root cause

Branch filenames were produced by replacing `/` with `-` and all other unsupported characters with `_`. That mapping is not one-to-one, so distinct valid refs such as `feature/auth` and `feature-auth` both resolved to `feature-auth.md`. Updates from either branch could therefore read, merge, or overwrite the other branch's memory.

## Compatibility

Existing branch state files are not renamed or deleted. Reads prefer the new hashed filename, then fall back to the legacy sanitized filename. The next state update reads the legacy content and writes it to the new collision-resistant path.

## Validation

- `npm run build` in `mcp/servers/egc-memory`
- `node tests/lib/branch-state.test.js` - 24 passed
- `node tests/scripts/egc-memory-branch-state.test.js` - 9 passed
- `node tests/scripts/watch-state.test.js` - 18 passed
- `node tests/hooks/claude-session-start.test.js` - 11 passed
- `node tests/hooks/egc-memory-save.test.js` - 12 passed
- `npm run lint` - 0 errors (existing warnings remain)
- `git diff --check`

---
## Summary by cubic
Prevent branch-state filename collisions by deriving filenames from a readable prefix plus a SHA-256 of the exact Git ref. Reads prefer the new hashed path with legacy fallback, and `egc watch` follows the same order.

- **Bug Fixes**
  - Introduced `branchStateKey` and write branch files as `<prefix>--<sha256>.md`.
  - Kept legacy sanitized filenames readable; reads fall back, next write migrates to the new path.
  - Updated `watch-state` to prefer hashed branch files and fall back to legacy.
  - Added regression tests for colliding branch names and legacy migration.

<sup>Written for commit 075a3d7356d8fc392e67bb7ae5173cf05dd2a3af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mcp/servers/egc-memory/src/branch-state.ts
+++ b/mcp/servers/egc-memory/src/branch-state.ts
@@ -1,7 +1,9 @@
 import path from 'node:path';
 import fs from 'node:fs';
+import { createHash } from 'node:crypto';
 
 export const DEFAULT_BRANCH_FILE = 'main.md';
+const BRANCH_FILE_PREFIX_LENGTH = 120;
 
 export type StateSource = 'branch' | 'default-branch' | 'flat' | 'none';
 
@@ -18,6 +20,15 @@ export function projectSlug(projectPath: string): string {
 
 export function sanitizeBranchName(branch: string): string {
   return branch.replaceAll('/', '-').replace(/[^a-zA-Z0-9-_]/g, '_');
+}
+
+// Sanitization alone is not injective: feature/auth and feature-auth both
+// become feature-auth. Keep a readable prefix, then bind it to the exact ref.
+export function branchStateKey(branch: string): string {
+  const branchName = String(branch || '');
+  const readablePrefix = sanitizeBranchName(branchName).slice(0, BRANCH_FILE_PREFIX_LENGTH) || 'branch';
+  const digest = createHash('sha256').update(branchName, 'utf8').digest('hex');
+  return `${readablePrefix}--${digest}`;
 }
 
 // Branch detection reads .git/HEAD instead of spawning git: no PATH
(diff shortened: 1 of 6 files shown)


Fmarzochi:
Welcome back @VIUK-XV, and what a return: your first code PR here, and it lands on a real, subtle bug. Great catch.

I reviewed the full diff and the design is exactly right: readable prefix bound to a SHA-256 of the exact ref, copy-on-write migration with legacy read fallback, TS and JS kept in sync, `egc watch` aligned, and solid regression coverage (the homedir mocking cleanup in the watch tests is a nice bonus, those used to touch the real `~/.egc`).

One non-blocking observation for a possible follow-up: with a 120-char prefix plus the 64-char digest, the filename can reach ~189 chars, which can brush against Windows MAX_PATH in deep home directories. Typical branch names stay around 100 chars so it's fine in practice, but a shorter prefix (say 40) or a truncated digest (16 hex is plenty for this use) would remove the edge entirely. Happy to take that as a tiny follow-up PR if you're interested, no need to change this one.

CI matrix is running; once it's green this gets merged.

VIUK-XV:

Fmarzochi:
Merged! 🎉 Welcome back to the contributors list, @VIUK-XV, this time with code!

This was a genuinely subtle bug (two different refs silently sharing one state file) and your fix is textbook: readable prefix bound to the exact ref, copy-on-write migration, watch alignment, regression tests, and even cleaning up the test hygiene along the way. The persistent-memory core of EGC is safer because of you.

You now hold both the Translator and Bug Hunter badges on the Hall of Fame. If you want to follow up, that MAX_PATH prefix trim we discussed is yours whenever you feel like it. ありがとうございました!

PR #912, opened 2026-07-19
fix(providers): raise NotImplementedError when stream=True

### LINKED ISSUE #903: fix(providers): stream flag silently ignored in the OpenAI-compatible flow
## Mission

Stop silently ignoring `stream=True` in the OpenAI-compatible provider flow.

## The bug

`LLMInput.stream` is accepted and propagated (for example `src/llm/providers/deepseek.py:80`), but the shared `OpenAIProvider.generate` (`src/llm/providers/openai.py:108`) never reads it when building the request params. Callers asking for streaming get a silent blocking call instead: the flag just vanishes.

## Where to look

- `src/llm/providers/openai.py:108` -- params construction ignores `llm_input.stream`
- `src/llm/providers/deepseek.py:80` -- one of the callers that forwards the flag

## Fix hint

Either implement streaming (yield chunks) or fail loudly with `NotImplementedError("streaming not supported")` when the flag is set. Silent downgrade is the only wrong option.

## Acceptance criteria

- [ ] `stream=True` either streams or raises explicitly
- [ ] A test covers the chosen behavior
- [ ] Non-streaming calls unchanged

## How to contribute

1. Fork the repo and create a branch for this issue
2. Sign your commits: `git commit -s` (DCO is enforced by CI; one-time setup: `git config core.hooksPath .githooks`)
3. The CLA bot will ask you to sign on your first PR: just follow its comment
4. Open a PR referencing this issue

First time here? Read [CONTRIBUTING.md](https://github.com/Fmarzochi/EGC/blob/main/.github/CONTRIBUTING.md). Questions? Ask right here in the issue or on [Discord](https://discord.gg/TxppsGb52): happy to mentor you through it.

Sufiyan-MSA:
## Summary
Closes #903.
 `LLMInput.stream` was accepted and propagated through the OpenAI-compatible flow but never read when `OpenAIProvider.generate` built the request params, so `stream=True` silently downgraded to a blocking call. Per the issue's acceptance criteria, this fails loudly with `NotImplementedError("streaming not supported")` instead.

## Changes
- `src/llm/providers/openai.py`: raise `NotImplementedError("streaming not supported")` at the top of `generate()` when `input.stream` is truthy. Non-streaming path unchanged.
- `src/llm/providers/deepseek.py`: add a narrow `except NotImplementedError: raise` so the contract error surfaces as-is instead of being caught by the broad `except Exception` that re-wraps native SDK errors as provider-attributed `LLMError`.

## Validation
- `python -m pytest tests/test_openai_provider.py tests/test_deepseek_provider.py` — 30 passed (27 pre-existing + 3 new).
- `python -m pytest tests/` — 117 passed, no regressions.
- New tests cover: (1) OpenAI `stream=True` raises `NotImplementedError` and the client is never called; (2) DeepSeek `stream=True` raises `NotImplementedError` (not wrapped as `LLMError`); (3) OpenAI `stream=False` behavior unchanged and no `stream` kwarg is forwarded to the SDK.

---
## Summary by cubic
Raise NotImplementedError("streaming not supported") when LLMInput.stream is True across OpenAI-compatible providers to prevent silent downgrade to a blocking call. DeepSeek, Groq, and Mistral now surface this contract error unchanged.

- Bug Fixes
  - OpenAIProvider.generate: early NotImplementedError when stream=True; non-stream path unchanged and no `stream` kwarg forwarded.
  - DeepSeekProvider, GroqProvider, MistralProvider: let NotImplementedError bubble (not wrapped as `LLMError`); tests cover stream=True errors with clients not called and add non-streaming regressions for OpenAI and DeepSeek.

<sup>Written for commit 9f24f7c1f3a11264ecb907af770e66a42609d418. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/llm/providers/deepseek.py
+++ b/src/llm/providers/deepseek.py
@@ -91,7 +91,7 @@ class DeepSeekProvider(OpenAIProvider):
             raise
         except NotImplementedError:
             # Contract-level errors (e.g. streaming not supported) are not
-            # provider-attributable wire failures — let them surface as-is.
+            # provider-attributable wire failures - let them surface as-is.
             raise
         except Exception as exc:
             # fix #2: native OpenAI SDK exceptions (RateLimitError,
--- a/src/llm/providers/deepseek.py
+++ b/src/llm/providers/deepseek.py
@@ -89,6 +89,10 @@ class DeepSeekProvider(OpenAIProvider):
             # plain LLMError here would discard that subclass information.
             exc.provider = ProviderType.DEEPSEEK
             raise
+        except NotImplementedError:
+            # Contract-level errors (e.g. streaming not supported) are not
+            # provider-attributable wire failures — let them surface as-is.
+            raise
         except Exception as exc:
             # fix #2: native OpenAI SDK exceptions (RateLimitError,
             # APIConnectionError, AuthenticationError, etc.) propagate here
--- a/src/llm/providers/groq.py
+++ b/src/llm/providers/groq.py
@@ -61,6 +61,10 @@ class GroqProvider(OpenAIProvider):
             # plain LLMError here would discard that subclass information.
             exc.provider = ProviderType.GROQ
             raise
+        except NotImplementedError:
+            # Contract-level errors (e.g. streaming not supported) are not
+            # provider-attributable wire failures - let them surface as-is.
+            raise
         except Exception as exc:
             # Native OpenAI SDK exceptions (RateLimitError, APIConnectionError,
             # AuthenticationError, etc.) propagate here unwrapped when
--- a/src/llm/providers/mistral.py
+++ b/src/llm/providers/mistral.py
@@ -58,6 +58,10 @@ class MistralProvider(OpenAIProvider):
             # plain LLMError here would discard that subclass information.
             exc.provider = ProviderType.MISTRAL
             raise
+        except NotImplementedError:
+            # Contract-level errors (e.g. streaming not supported) are not
+            # provider-attributable wire failures - let them surface as-is.
+            raise
         except Exception as exc:
             # Native OpenAI SDK exceptions (RateLimitError, APIConnectionError,
             # AuthenticationError, etc.) propagate here unwrapped when
--- a/src/llm/providers/openai.py
+++ b/src/llm/providers/openai.py
@@ -95,6 +95,11 @@ class OpenAIProvider(LLMProvider):
         raise e
 
     def generate(self, input: LLMInput) -> LLMOutput:
+        if input.stream:
+            # Streaming is not implemented in this adapter. Fail loudly instead
+            # of silently downgrading to a blocking call, which would mislead
+            # callers into thinking they are consuming a stream.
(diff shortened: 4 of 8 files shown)


Sufiyan-MSA:

Sufiyan-MSA:
Hi @Fmarzochi — CLA Assistant Lite has confirmed "All contributors have signed the CLA" but the CLA / cla workflow check is still red with "Committers of Pull Request 912 have to sign the CLA". Looks like the two CLA systems on the repo are not in sync. Could you take a look when you get a chance? Signature is on file via the Lite bot.

Fmarzochi:
Hi @Sufiyan-MSA, thanks for the quick turnaround on this one. About the CLA: the check was stuck on a stale run from before your signature. I re-ran it and it is green now, so nothing else is needed on that front.

On the code: the fix is correct and well tested for OpenAI and DeepSeek, but there is one gap to close before merging. `GroqProvider` and `MistralProvider` also subclass `OpenAIProvider` and wrap `super().generate()` in the same `try/except Exception` block that DeepSeek has. Your guard in `OpenAIProvider.generate()` now raises for them too, but without the `except NotImplementedError: raise` bypass their wrappers will swallow the error and re-raise it as a generic `LLMError`, which masks exactly the contract failure you worked to surface. (`OpenRouterProvider` is fine: it inherits `generate()` without wrapping.)

Could you:

1. Add the same `except NotImplementedError: raise` bypass to `groq.py` and `mistral.py`, with matching `test_stream_flag_raises_not_implemented` tests for both providers.
2. Add a `test_non_streaming_call_unchanged` test to `test_deepseek_provider.py`, mirroring the one you wrote for OpenAI, so both providers pin the stream=False path.
3. Tiny style nit: replace the em dash in the new deepseek.py comment with a plain hyphen. We keep the codebase free of em dashes.

Everything else looks solid: the check sits at the right layer (before params are built, client never called), it fails loudly instead of silently downgrading, and the mock assertions are exactly what the issue asked for. Once Groq and Mistral are covered this is ready to go.

Fmarzochi:
Merging. This closed the loop exactly right: the guard sits in the shared flow, all three wrapper subclasses now re-raise NotImplementedError instead of masking it as a generic LLMError, and every provider got both the stream=True raise test and the stream=False regression pin. Thanks for taking the review feedback and covering Groq and Mistral in one clean pass, @Sufiyan-MSA. Welcome aboard, your Hall of Fame entry is coming right up. If you want a natural follow-up, #915 proposes deduplicating exactly the wrapper blocks you just touched.

Fmarzochi (approved):

Issue #987, opened 2026-07-23
Break EGC on purpose: debug the whole runtime, every provider, every promise


Fmarzochi:
## Who I am looking for

I specifically want a tester running EGC on the **Mistral** model or **Anthropic's Fable** (any other backend is welcome too: Claude, GPT-4o, Gemini, DeepSeek, Groq, Cohere, Vertex AI, OpenRouter for Qwen3 and Llama 4). Provider-specific behavior is exactly what I want stressed, so run it on Mistral or Fable first and tell me what breaks.

## The mission

EGC promises a lot in the README. Prove or break every single claim. Install it, then push the whole runtime until something cracks: persistent memory, the Guardian safety layer, the Token Crusher, the dashboard, the cross-harness install, every provider. If the README promises it, test it. If it does not do what it says, that is exactly what I want to know.

This is a deep debugging pass, not a five-minute smoke test. Read the code, run it for real, and hunt.

## Get the debug build (run it from source)

To literally step through the code, install from source instead of the published package:

```bash
git clone https://github.com/Fmarzochi/EGC
cd EGC
sh scripts/install.sh
egc doctor
```

Prefer the one-liner? `npm install -g @egchq/egc && egc install` works too, but the source checkout lets you read and instrument everything.

## What to hammer (every promise in the README)

- **Run it on your model first: Mistral or Anthropic Fable.** Point EGC at that backend and watch for provider-specific breakage before anything else, then repeat on the others.
- **Cross-harness install** across the 23 supported tools (Cursor, Claude Code, Codex, Copilot, Aider, Gemini CLI, and the rest). Does `egc install` register the MCP servers everywhere it claims, and does `egc doctor` report the truth?
- **Persistent memory** (`get_state` / `update_state`). Start a session, save state, kill it, restore it. Does the brain actually carry over? Try it across parallel sessions too.
- **Guardian** safety layer. Does `validate_command` block dangerous commands, does `validate_write` stop dangerous writes, and does prompt-injection detection actually fire?
- **Token Crusher.** Run noisy commands (git logs, test spam, huge JSONs). Is the compression real (up to 90%) and does it preserve every error and warning? Ask "how much did I save?" in any language and check the number against the local ledger (`egc gain` / `egc saved` / `egc discover`).
- **Dashboard.** Does the live dashboard spin up after install and show real agent activity, tokens, and cost?
[...]

PR #1001, opened 2026-07-25
sec: guardian security limits and command allowlist expansion

fuentes71:
Expands the guardian command allowlist to match the real daily workflow and ships the pending security-limit work on this branch.

Allowlist (new commit):
1. SAFE_DEV gains egc, multica, gh, docker, prisma, php, pnpm, turbo. SAFE_READONLY gains dir, where, where.exe. These were all hitting a blanket deny (see ~/.egc/audit.log full of DENIED since 2026-07-24), which made validate_command useless in practice.
2. New CLIs are not blanket-trusted: docker blocks prune/rm/rmi and down with volumes, gh blocks delete, prisma blocks reset and --force-reset.
3. powershell, wsl and Remove-Item stay denied: the first two are wrappers that would smuggle arbitrary commands past validation, the last is rm.

Pre-existing commits on this branch, included here for review: SEC-04/05/06 and REL-06 path validation and limits, two SonarQube fixes, one CI unicode fix.

---
## Summary by cubic
Expands the guardian allowlist to support daily tooling while blocking destructive flags, refactors validation into a per-command dispatch map, and enforces strict path and size limits for context and `project_path`. Implements SEC-04/05/06 and REL-06 path validation and limits.

- **New Features**
  - Expanded allowlist: added `egc`, `gh`, `docker`, `prisma`, `php`, `pnpm`, `turbo`; read-only adds `dir`, `where`, `where.exe`. Removed `multica`.
  - Destructive-arg guards: block `docker prune/rm/rmi`, `compose down -v/--volumes`, `docker run/create` with host mounts or `--privileged`; block `gh` deletes (including `-X/--method DELETE`); block Prisma `reset`/`--force-reset`, `db execute`, and `--accept-data-loss`. `powershell`, `wsl`, and `Remove-Item` remain denied.
  - Path and size limits: protect `~/.egc`; context files must be regular files, ≤10MB each and ≤50MB total; realpath and validate `project_path`, reject protected paths.

- **Refactors**
  - Validator reworked to a per-command dispatch map; added a validator test covering allow/deny sets. Also improved error logging for Windows hidden-attr setup and persistent logger writes, and made CI treat crash-with-no-output as one unknown failure.

<sup>Written for commit f045ab25906a6517d8ef1e6ce36584e0a1e4e420. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1001?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Security**
  - Strengthened filesystem protections by verifying resolved paths are regular files, enforcing per-file and total size limits, and recording denials for oversized or invalid inputs.
  - Expanded command allowlists and tightened protections against destructive actions (e.g., Docker, GitHub, Prisma), including additional protected locations.
  - Improved validation for learning operations by resolving and validating real project paths before proceeding.
- **Reliability**
  - Added explicit non-critical error logging for Windows root-hiding and improved visibility when audit logging fails.
  - Improved test runner messaging for crashes with no output.
- **Behavior Changes**
  - Task routing now relies on the request prompt and provided file paths only.

### DIFF
--- a/mcp/servers/egc-guardian/src/index.ts
+++ b/mcp/servers/egc-guardian/src/index.ts
@@ -343,8 +343,6 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
 
       case "orchestrate_task": {
         const prompt = request.params.arguments?.prompt as string;
-        const agents: string[] = (request.params.arguments?.agents ?? []) as string[];
-        const skills: string[] = (request.params.arguments?.skills ?? []) as string[];
         const files: string[] = (request.params.arguments?.filepaths ?? []) as string[];
 
         // Read any provided file payloads
--- a/mcp/servers/egc-guardian/src/index.ts
+++ b/mcp/servers/egc-guardian/src/index.ts
@@ -13,8 +13,8 @@ function hideEgcRootOnWindows(): void {
   const egcRoot = path.join(os.homedir(), '.egc');
   try {
     execSync(`attrib +h "${egcRoot}"`, { stdio: 'ignore' });
-  } catch (_) {
-    // non-critical: folder works even if attribute fails
+  } catch (error) {
+    console.error('non-critical: folder works even if attribute fails', error);
   }
 }
 import { z } from 'zod';
--- a/mcp/servers/egc-guardian/src/index.ts
+++ b/mcp/servers/egc-guardian/src/index.ts
@@ -106,7 +106,9 @@ class PersistentLogger {
         // file might not exist
       }
       await fs.promises.appendFile(this.logPath, payload + '\n', 'utf-8');
-    } catch(e) {}
+    } catch (e) {
+      console.error('[EGC guardian] Log write failed (disk full?):', String(e));
+    }
   }
 }
 
--- a/mcp/servers/egc-guardian/src/index.ts
+++ b/mcp/servers/egc-guardian/src/index.ts
@@ -267,6 +267,19 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
                auditLog('CONTEXT_LOAD', 'DENIED', { filepath, reason: 'protected path' });
                continue;
              }
+             const stats = fs.statSync(realResolved);
+             if (!stats.isFile()) {
+               auditLog('CONTEXT_LOAD', 'DENIED', { filepath, reason: 'not a regular file' });
+               continue;
+             }
+             if (stats.size > 10 * 1024 * 1024) { // 10MB per file max
+               auditLog('CONTEXT_LOAD', 'DENIED', { filepath, reason: 'exceeds per-file 10MB limit' });
+               continue;
+             }
+             if (totalBytesLoaded + stats.size > 50 * 1024 * 1024) { // 50MB aggregate max
+               auditLog('CONTEXT_LOAD', 'DENIED', { filepath, reason: 'exceeds aggregate 50MB limit' });
+               continue;
+             }
              const content = await fs.promises.readFile(realResolved, 'utf-8');
              // Split context into chunks/paragraphs to allow granular pruning
              const chunks = content.split('\n\n').filter(c => c.trim().length > 0);
@@ -374,6 +387,15 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
 
         if (!projectPath) {
           throw new McpError(ErrorCode.InvalidParams, 'project_path is required');
+        }
+        let realProjectPath: string;
+        try {
+          realProjectPath = fs.realpathSync(path.resolve(projectPath));
+        } catch {
+          throw new McpError(ErrorCode.InvalidParams, 'project_path resolution failed');
+        }
+        if (isProtectedPath(realProjectPath)) {
+          throw new McpError(ErrorCode.InvalidParams, 'project_path must not be a protected path');
         }
 
         const result = await autoLearn({ project_path: projectPath, target_file: targetFile, limit });
--- a/mcp/servers/egc-guardian/src/index.ts
+++ b/mcp/servers/egc-guardian/src/index.ts
@@ -14,7 +14,7 @@ function hideEgcRootOnWindows(): void {
   try {
     execSync(`attrib +h "${egcRoot}"`, { stdio: 'ignore' });
   } catch (error) {
-    console.error('non-critical: folder works even if attribute fails', error);
+    console.error('non-critical: folder works even if attribute fails');
   }
 }
 import { z } from 'zod';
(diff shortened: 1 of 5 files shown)


Fmarzochi (changes requested):
Thanks Matheus, the SEC/REL limits work is in the right direction and I want most of it in. But I can't take the PR as it stands. Grouped feedback:

**Blockers**

1. `multica` has to come out of `SAFE_DEV` (and out of the switch case). That's internal tooling from my private stack, not a public CLI this product should know about; the shipped allowlist stays limited to broadly known dev tools. Please update the PR description on the next push as well. If your local workflow needs extra commands, that's exactly the case for a user-extensible allowlist policy (config based), which I'd rather track as a separate follow-up issue than solve with hardcoded names.

2. `dir`, `where` and `where.exe` were added to `SAFE_READONLY` but have no case in `validateCommandArgs`, so they fall through to the `default` branch and come back allowed without the per-argument `isProtectedPath` check that `ls` and `cat` do. As is, `dir ~/.ssh` or `where /r <protected dir>` walks straight past the guard. They need the same protected-path argument handling. Good catch adding the Windows utilities at all, by the way; the repo rule is that Windows is never left out, it just needs the same rigor.

3. The new `SAFE_DEV` entries are much broader than their guards:
   - `gh`: `args.includes('delete')` misses `gh api -X DELETE ...` and `--method DELETE`, and `gh alias set` / `gh extension exec` are arbitrary execution under the CLI's token.
   - `docker`: `run`, `exec` and `cp` remain fully allowed; `docker run -v /:/host --privileged ...` is root on the host. Blocking `rm/rmi/prune/down -v` does not contain that.
   - `prisma`: `db execute` (arbitrary SQL) and `db push --accept-data-loss` sail past the `reset` guard.
   - `php -r`, `pnpm dlx` / `pnpm exec` and `turbo run` are arbitrary code execution by design.
   - The parsing also misses `--flag=value` forms (`--volumes=true`, `--method=DELETE`).

   Either scope these CLIs to explicit subcommand allowlists (for example `docker build` / `docker compose ps`, `gh pr` / `gh issue` / `gh api` restricted to GET) or leave them out of this PR and let the config-based policy cover per-user needs. A hardcoded arms race of deny flags won't hold.

4. Consistency of the limits themselves:
   - The new isFile + 10MB per file / 50MB aggregate checks only cover `reduce_context`; `orchestrate_task` still reads `filepaths` unchecked.
   - `auto_learn` now validates `project_path` but not `target_file`, which can still point into protected paths.
   - In egc-memory, `resolveProjectPath` only rejects `..`; it should also go through `isProtectedPath`.

**Also needed before re-review**

- SonarCloud Quality Gate is at Maintainability D on new code (required >= A) and the CI step `Validate Components / Check unicode safety` is failing on the branch. Both green, please.

**Suggestion, not blocking**

- The PR mixes the allowlist expansion with the SEC-04/05/06 + REL-06 limits. If you split them, the limits half can land much faster while we settle the allowlist policy.

The limits are genuinely useful work. Trim the allowlist to what the guards can honestly contain, fix the read bypass on the Windows utilities, and this moves quickly.

Fmarzochi:
Two pointers to speed up the green run:

1. The Quality Gate D comes from function size and complexity in the new-code scope: validateCommandArgs is now about 160 lines with a 14-case switch. A small dispatch map (command to validator function, each one 10-15 lines) brings it back under the cognitive-complexity threshold and reads better. The long index.ts handlers your commits touch benefit from the same extraction.

2. The unicode step: scripts/ci/check-unicode-safety.js passes clean on your branch locally (1581 files checked), so the CI failure is environmental. Open the failing job log and look for the 'Unicode safety violations detected: file:line:col U+XXXX' line to get the exact file; artifacts generated only in CI belong in the script's ignoredDirs, real source violations need the character removed.

Fmarzochi:
Matheus, I pushed a commit finishing this off so we can land it, keeping your co-authorship. What changed and why:

- Refactored `validateCommandArgs` from the 14-case switch into a per-command dispatch map. That switch (~160 lines) was the source of the Maintainability D on new code; each validator is now small and single-purpose, so the Sonar gate should clear.
- Removed `multica` from the allowlist. It's internal tooling from my private stack, not a public CLI this product ships, so it doesn't belong in the shared allowlist. If your local flow needs extra commands, a config-based user allowlist is the right home for that, as a separate follow-up.
- `dir`, `where` and `where.exe` were in `SAFE_READONLY` but had no case, so they fell through to `default` and skipped the protected-path check: `dir ~/.ssh` walked straight through. They now share the read-only path validator with `ls`/`cat`.
- Hardened the CLIs you added, keeping normal use open: `docker run/create` with a host mount (`-v`, `--mount`) or `--privileged`/`--cap-add`, `gh` delete passed as `-X`/`--method DELETE`, and `prisma db execute` / `--accept-data-loss` are now blocked; `docker build`/`compose up`, `gh pr list`, `prisma generate` stay allowed.
- Added `tests/egc-guardian-validator.test.js` covering the allow and deny sets.

The SEC-04/05/06 and REL-06 limits are the valuable core here and they stay untouched. Thanks for driving this.

Fmarzochi:
Matheus, I dug into merging this and hit a real blocker: the branch is based on a commit from well before a large batch of validator and MCP-server work already on main. The command validator was rewritten upstream to block inline-eval flags (`node -e`, `php -r`, `bash -c`), `find -delete`/`-exec`, embedded `--file=` paths, and an expanded protected-file set covering `.bashrc`, `.gitconfig` and each tool's credential files, plus the session-bus and #987 runtime-bug fixes. Rebasing this branch conflicts across ~2000 lines in the three core files, and a straight merge would roll those protections back.

So I can't take it as-is without regressing security that's already shipped. Two clean paths:

1. Rebase `sec/guardian-limits` onto current `main` and keep only what main doesn't already have: the SEC-04/05/06 + REL-06 path/size limits. Most of the allowlist and validator hardening here is already covered upstream, so the diff should shrink a lot.
2. If that's more churn than it's worth, I can lift just the limits into a fresh PR off main and credit you.

Either way the limits idea is good and worth landing. Tell me which you prefer and I'll help get it in.

Fmarzochi:
Matheus, following up on the two paths I laid out above. I checked what each one would actually keep, and the honest answer changes the plan.

The pieces worth landing here already went in on main through a later batch: the per-file and aggregate size caps on `reduce_context` live in the guardian now, and `resolveProjectPath` already resolves the real path and runs it through the protected-path check. So a rebase of this branch would come back nearly empty, not because your reasoning was off but because we converged on the same protections in parallel while this branch sat on an older base.

Given that, I'm going to close this one to keep the tree tidy. The thinking was sound end to end, and the review it triggered was useful.

The one idea here that is not upstream yet is extending the allowlist for local tooling. That belongs as a config-based user allowlist rather than hard-coded entries, and it is worth doing. If you want it, say the word and I'll open a fresh issue off main with the shape of it, assigned to you. Thanks for pushing on this.

PR #1089, opened 2026-07-29
ci(fuzzing): complete fork bypass in cflite_pr.yml, pin revert SHA in all workflows

Fmarzochi:
Follow-up to #1075.

Three fixes after the CIFuzz fork merge:

1. cflite_pr.yml was not updated in #1075 -- now points to Maqbool61/clusterfuzzlite@f3fd422 with sanitizer: none and language: javascript, matching batch and cron.

2. Revert comment in all three workflow files referenced `@v1` (floating tag). Now references the reviewed pinned SHA `884713a6c30a92e5e8544c39945cd7cb630abcd1`, so a future revert stays reproducible.

3. Permissions confirmed at minimum: `contents: read` + `security-events: write` (required for SARIF upload). No scope reduction possible without splitting jobs.

---
## Summary by cubic
Completes the ClusterFuzzLite fork bypass and pins revert SHAs to keep CI fuzzing stable and reproducible. Aligns `cflite_pr.yml` with batch/cron using `sanitizer: none` for JavaScript and minimal permissions.

- **Bug Fixes**
  - Switches `cflite_pr.yml` to `Maqbool61/clusterfuzzlite@f3fd422fd0d55f2b4b2c3c1795ed4bcb7084ca41` for build/run; sets `language: javascript` and `sanitizer: none` to match batch/cron.
  - Replaces floating revert hint to `google/clusterfuzzlite/actions/build_fuzzers@v1` with pinned `@884713a6c30a92e5e8544c39945cd7cb630abcd1` in all workflows.
  - Confirms minimum permissions for SARIF upload: `contents: read`, `security-events: write`.

<sup>Written for commit 5a99e73155e4392c958981330dd0d686f79c6148. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1089?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/cflite_batch.yml
+++ b/.github/workflows/cflite_batch.yml
@@ -25,7 +25,7 @@ jobs:
         id: build
         # Pinned to Maqbool61/clusterfuzzlite@fix/js-sanitizer-none (EGC#910):
         # patches config_utils.py inside the Docker image to allow SANITIZER=none
-        # for JavaScript projects. Revert to google/clusterfuzzlite/actions/build_fuzzers@v1
+        # for JavaScript projects. Revert to google/clusterfuzzlite/actions/build_fuzzers@884713a6c30a92e5e8544c39945cd7cb630abcd1
         # once google/oss-fuzz#15907 is resolved upstream.
         uses: Maqbool61/clusterfuzzlite/actions/build_fuzzers@f3fd422fd0d55f2b4b2c3c1795ed4bcb7084ca41
         with:
--- a/.github/workflows/cflite_cron.yml
+++ b/.github/workflows/cflite_cron.yml
@@ -25,7 +25,7 @@ jobs:
         id: build
         # Pinned to Maqbool61/clusterfuzzlite@fix/js-sanitizer-none (EGC#910):
         # patches config_utils.py inside the Docker image to allow SANITIZER=none
-        # for JavaScript projects. Revert to google/clusterfuzzlite/actions/build_fuzzers@v1
+        # for JavaScript projects. Revert to google/clusterfuzzlite/actions/build_fuzzers@884713a6c30a92e5e8544c39945cd7cb630abcd1
         # once google/oss-fuzz#15907 is resolved upstream.
         uses: Maqbool61/clusterfuzzlite/actions/build_fuzzers@f3fd422fd0d55f2b4b2c3c1795ed4bcb7084ca41
         with:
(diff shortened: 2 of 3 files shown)


Fmarzochi:
Thanks for the P1 flag, @cubic-dev-ai — the concern is valid and was evaluated before merge. Here is the full breakdown.

**What the fork actually executes**

The diff between `google/clusterfuzzlite@884713a6c30a92e5e8544c39945cd7cb630abcd1` and `Maqbool61/clusterfuzzlite@f3fd422fd0d55f2b4b2c3c1795ed4bcb7084ca41` restructures both action files from `using: 'docker'` to `using: 'composite'`, but **does not introduce a custom Docker image**. The container used in both steps is still `gcr.io/oss-fuzz-base/clusterfuzzlite-build-fuzzers:v1` — the same image Google ships. The only custom code is a single `sed` command that runs inside that container:

```bash
sed -i "s/SANITIZERS = \['address', 'memory', 'undefined', 'coverage'\]/SANITIZERS = ['address', 'memory', 'undefined', 'coverage', 'none']/" \
  /opt/oss-fuzz/infra/cifuzz/config_utils.py
```

This patch fires only when `language == 'javascript' && sanitizer == 'none'`. All other language/sanitizer combinations fall through to the original Docker path unchanged.

**Why SHA-pinning is sufficient here**

The SHA pin (`f3fd422fd0d55f2b4b2c3c1795ed4bcb7084ca41`) was verified against the above diff before merge. It guarantees we execute exactly the reviewed commit. The fork cannot drift past this point.

**Scope of `GITHUB_TOKEN` and `security-events: write`**

Passing `GITHUB_TOKEN` to the container is identical behavior to the upstream Google action — the original `using: 'docker'` approach exposes the token to the entrypoint in the same way. The `security-events: write` permission is required exclusively by the `github/codeql-action/upload-sarif` step, which uses a trusted Google-maintained action and is separate from the fork steps.

Additionally, all three `cflite_*.yml` workflows are gated to `workflow_dispatch` only — they do not execute automatically on pull requests, which limits the blast radius if the assessment above were ever wrong.

**Temporary measure**

This workaround will be reverted to the upstream Google action once [google/oss-fuzz#15907](https://github.com/google/oss-fuzz/issues/15907) is resolved. The revert target SHA and conditions are documented in the workflow comments.

PR #1102, opened 2026-07-30
fix(memory): stop propagate.ts from discarding real content in .cursor context file

Fmarzochi:
## Summary
- writeCursorContext's legacy-migration heuristic in propagate.ts stripped everything after the old unmarked frontmatter in .cursor/rules/egc-context.mdc whenever no `<!-- egc:start -->` marker existed yet, including content a human had added there by hand, not just the pre-marker writer's own auto-generated block.
- Now it only strips when what follows the frontmatter is literally the old auto-generated block header ("## EGC Project Memory"), and leaves any other content untouched so the marked EGC block gets appended below it instead of replacing it.
- This is one of the causes behind update_state / propagateStateToTools silently discarding real content in the .cursor context file on every call.

## Test plan
- [x] Isolated Node script against the built module (not the MCP server) covering three scenarios: (1) real hand-written content after the legacy frontmatter survives two consecutive propagate calls, (2) a purely auto-generated legacy leftover still migrates cleanly without duplicating the block, (3) AGENTS.md-style files with real content survive unchanged. All 9 checks pass with the fix; the same script reproduces the original data-loss bug without it.

---
## Summary by cubic
Stops both propagate paths from discarding real content in `.cursor/rules/egc-context.mdc` when migrating legacy files without `<!-- egc:start -->`; only the old auto-generated block is removed and a marked EGC block is appended (EGC-513).

- **Bug Fixes**
  - In both `mcp/servers/egc-memory/src/propagate.ts` and `scripts/lib/propagate-state.js`, strip after the legacy frontmatter only when the next line is "## EGC Project Memory"; otherwise keep content and append the marked block via `upsertEgcSection`.
  - Added tests to cover preservation, clean migration (no duplicates), and skipping writes when `.cursor/` is absent.

<sup>Written for commit ab9033b9fa8b85f07829a9d74caea1f675fb40dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mcp/servers/egc-memory/src/propagate.ts
+++ b/mcp/servers/egc-memory/src/propagate.ts
@@ -82,9 +82,14 @@ function upsertFileSection(filePath: string, block: string, defaultContent = '')
 // below it -- permanent duplication. Recognize and strip it down to just
 // the frontmatter so only the new marked block survives.
 const LEGACY_CURSOR_FRONTMATTER = `---\ndescription: EGC project memory (auto-updated by update_state)\nalwaysApply: true\n---\n\n`;
+const LEGACY_BLOCK_HEADER = '## EGC Project Memory';
 function stripLegacyCursorContent(existing: string): string {
   if (existing.includes(EGC_START)) return existing;
-  return existing.startsWith(LEGACY_CURSOR_FRONTMATTER) ? LEGACY_CURSOR_FRONTMATTER : existing;
+  if (!existing.startsWith(LEGACY_CURSOR_FRONTMATTER)) return existing;
+  // Only the pre-marker writer's own auto-generated block is safe to drop here.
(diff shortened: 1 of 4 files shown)


Fmarzochi:
Reported directly by Helal Ferrari Cabral (@helalferrari), who is not a GitHub contributor on this repo but flagged the memory corruption behavior that this PR fixes. Thank you.

PR #1103, opened 2026-07-30
fix(memory): stop upsertEgcSection from deleting content on orphaned markers

Fmarzochi:
## Summary
- Squad review (EGC-514) of the previous EGC-513 fix found a more serious variant of the same bug: `upsertEgcSection` itself -- shared by all 14 propagated files (AGENTS.md, GEMINI.md, CLAUDE.md, .cursor, etc.), not just the .cursor migration path -- could permanently delete real content if `<!-- egc:end -->` was ever removed or corrupted while `<!-- egc:start -->` survived. The call after that would slice-and-replace everything between an orphaned start marker and a freshly appended end marker.
- `upsertEgcSection` (in both `mcp/servers/egc-memory/src/propagate.ts` and `scripts/lib/propagate-state.js`) now only does the in-place slice when there is exactly one correctly paired start/end marker set. Any other shape (missing, orphaned, duplicated, or inverted markers) strips just the bare marker tags and appends one fresh, correctly paired block -- it never slices on marker positions it can't be sure are matched.
- Also normalizes CRLF before comparing against the legacy cursor frontmatter, so files saved with Windows line endings are recognized and migrated instead of silently duplicating the block.

## Test plan
- [x] 4 new scenarios added to both test suites: orphaned end marker survives two consecutive calls, duplicated marker pairs collapse to one without losing real content, inverted marker order doesn't delete real content, CRLF-saved legacy cursor file still migrates cleanly.
- [x] `tests/scripts/egc-memory-propagate.test.js`: 24/24 passing locally.
- [x] `tests/scripts/propagate-state.test.js`: 33/33 passing locally.

---
## Summary by cubic
Prevents data loss by making `upsertEgcSection` safe when EGC markers are broken across all propagated files. Addresses EGC-513 (found during EGC-514 review) and ensures CRLF legacy `.cursor` files migrate without duplicate blocks.

- **Bug Fixes**
  - `upsertEgcSection` now slices only when there is exactly one matched start/end pair; for missing, orphaned, duplicated, or inverted markers it strips bare tags and appends a fresh paired block.
  - Normalizes CRLF before checking legacy `.cursor` frontmatter so Windows-saved files are recognized and migrated cleanly.
  - Excludes the `scripts/lib/propagate-state.js` and `mcp/servers/egc-memory/src/propagate.ts` mirror pair from Sonar duplication to avoid false positives; no runtime changes.

<sup>Written for commit a8a52b78a92a4801e4571340c3b2a5e29353899e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -6,11 +6,14 @@ sonar.sources=src,mcp/servers,scripts
 # These tests run in CI on every push; CI remains their quality gate.
 sonar.exclusions=**/node_modules/**,**/dist/**,**/coverage/**,**/*.tgz,**/legacy-command-shims/**
 sonar.sourceEncoding=UTF-8
-# branch-state and global-state are deliberate CommonJS mirrors of their TS
-# counterparts: session hooks cannot require the compiled server build.
-# tests/egc-memory-global-state.test.js asserts the global-state pair stays
-# behaviorally identical.
-sonar.cpd.exclusions=scripts/lib/branch-state.js,mcp/servers/egc-memory/src/branch-state.ts,mcp/servers/egc-guardian/src/catalog-index.ts,scripts/lib/global-state.js,mcp/servers/egc-memory/src/global-state.ts
+# branch-state, global-state and propagate-state are deliberate CommonJS
+# mirrors of their TS counterparts: session hooks and the CLI cannot require
+# the compiled MCP server build. tests/egc-memory-global-state.test.js
+# asserts the global-state pair stays behaviorally identical; the
+# propagate-state pair has matching scenarios across
+# tests/scripts/propagate-state.test.js and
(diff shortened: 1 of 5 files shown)


Fmarzochi:
Same report from Helal Ferrari Cabral (@helalferrari) as #1102: he caught the orphaned-marker bug that was silently deleting real content in propagated files. Thank you.

PR #1104, opened 2026-07-30
fix(cli): stop bare terminal invocations from misrouting to an installed harness dir (BUG-08)

Fmarzochi:
## Summary
- `getEGCDir()`'s Tier 3 fallback picked "the first harness dir that exists under HOME" whenever no CLI-specific env var or `__dirname` match applied (e.g. any plain `egc <command>` run from a terminal). On a machine with OpenCode installed, that silently pointed doctor/state-store calls at `~/.config/opencode/egc/state.db` instead of the shared `~/.egc` store, splitting state across databases that never saw each other's writes -- and making `egc doctor`'s own divergent-database warning unreliable, since it depends on both paths resolving under the same root.
- Tier 3 now checks the tool-agnostic `~/.egc` default first and only falls back to "first installed harness dir" when `~/.egc` doesn't exist yet, preserving the original dev/test intent for that narrower case.

## Test plan
- [x] 2 new scenarios in `tests/lib/utils.test.js`: `~/.egc` wins over an installed harness dir; still falls back to a harness dir when `~/.egc` doesn't exist yet.
- [x] `tests/lib/utils.test.js`: 185/185 passing locally.
- [x] `tests/scripts/doctor.test.js`: 11/11 passing locally (including the existing divergent-database test).
- [x] Manually confirmed `node scripts/doctor.js` now correctly reports the divergent-database warning from a bare terminal invocation.

---
## Summary by cubic
Fix CLI path resolution so bare terminal `egc` runs use the shared `~/.egc` state store, not a harness dir, and refactor `getEGCDir()` into small helpers to reduce complexity. This prevents split state and keeps `egc doctor`'s divergent-database check reliable (EGC-513, BUG-08).

- **Bug Fixes**
  - Updated `getEGCDir()` Tier 3: prefer `~/.egc` if it exists; only fall back to the first installed harness dir when `~/.egc` is missing.
  - Added two tests to verify precedence and fallback behavior.

<sup>Written for commit 54836ce54284958e9b7e6535f2661c05051c4555. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/lib/utils.js
+++ b/scripts/lib/utils.js
@@ -40,56 +40,21 @@ function getClaudeDir() {
   return getEGCDir();
 }
 
-/**
- * Get the EGC config directory for the active harness.
- *
- * Resolution order:
- *   1. EGC_DIR env var (explicit override)
- *   2. Harness-specific env vars injected at hook time (see docs/spec/harness-env-vars.md)
- *   3. __dirname prefix match against known harness home dirs (production install)
- *   4. ~/.egc (harness-agnostic fallback)
- */
-function getEGCDir() {
-  if (process.env.EGC_DIR) return process.env.EGC_DIR;
+// Tier 1: harness-specific env vars injected at hook time (see docs/spec/harness-env-vars.md).
+// Gemini CLI is checked before Claude Code because Gemini CLI sets both as a compat alias.
+function resolveHarnessDirFromEnv(env, home) {
+  if (env.GEMINI_PROJECT_DIR || env.GEMINI_PLUGIN_ROOT) return path.join(home, '.gemini');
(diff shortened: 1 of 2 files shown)


Fmarzochi:
Reported directly by Helal Ferrari Cabral (@helalferrari): the state.db divergence between harnesses that this PR fixes. Thank you.

PR #1130, opened 2026-08-02
fix(guardian): resolve DENIED_PATHS entries at comparison time, not load time

Fmarzochi:
## Summary
- buildDeniedPaths() resolved symlinks once at module load, creating a staleness window: a denied directory that becomes a symlink (or whose target starts existing) after the guardian process starts would bypass protection until a restart.
- Both the incoming path and each DENIED_PATHS entry now resolve through the shared resolveRealOrLexical() helper on every isProtectedPath() call, so there is no staleness window.
- Excludes fuzz/validator.cjs from ESLint (generated fuzz harness, not hand-written source).

Flagged in cubic review on PR #1129 as a follow-up to the EGC-538 macOS /etc symlink bug.

## Test plan
- [x] npm run build (egc-guardian package)
- [x] node tests/scripts/egc-guardian.test.js -- 125/125 passed locally
- [x] npx eslint on touched files -- 0 errors
- [ ] Full CI matrix (cloud)

---
## Summary by cubic
Resolve denied paths at comparison time in `egc-guardian` to remove the staleness window where new or retargeted symlinks could bypass protection (EGC-538/macOS /etc). Also remove a circular `node_modules` symlink from `egc-memory` and ignore a generated fuzz harness in ESLint.

- **Bug Fixes**
  - Guardian: Both input and denied paths now resolve via `resolveRealOrLexical()` on every check, preventing stale `DENIED_PATHS` and blocking symlink escapes (EGC-538).
  - Memory: Untracked circular `node_modules` symlink in `mcp/servers/egc-memory` to avoid broken clones.

- **Misc**
  - Tests now include end-to-end coverage for symlinked `DENIED_PATHS` and mid-run symlink retargets.
  - ESLint now ignores `fuzz/validator.cjs` (generated).

<sup>Written for commit f6348975ab6b08b4647e6692e2cded39982358fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -10,7 +10,8 @@ module.exports = [
             '.claude/**',
             'dashboard/**',
             'node_modules/**',
-            'mcp/servers/*/build/**'
+            'mcp/servers/*/build/**',
+            'fuzz/validator.cjs'
         ]
     },
     js.configs.recommended,
--- a/mcp/servers/egc-guardian/src/validator.ts
+++ b/mcp/servers/egc-guardian/src/validator.ts
@@ -702,25 +702,31 @@ export function buildDeniedPaths(): string[] {
     );
   }
 
-  // isProtectedPath() resolves the incoming path through fs.realpathSync()
-  // before comparing it against this list, so an entry that is itself a
-  // symlink must be resolved here too, or the comparison never matches.
-  // Concretely: on macOS /etc is a symlink to /private/etc, so a write to
-  // /etc/hosts or /etc/shadow resolves to /private/etc/hosts, which never
-  // starts with the literal '/etc' string below, silently defeating this
-  // protection on macOS only (audit EGC-538 sub-part 1, caught by CI).
-  const resolved = paths.map(p => {
-    try {
-      return fs.realpathSync(p);
-    } catch {
-      return p;
-    }
-  });
-
-  return [...new Set([...paths, ...resolved])].filter(Boolean);
+  return paths.filter(Boolean);
 }
 
 export const DENIED_PATHS: string[] = buildDeniedPaths();
+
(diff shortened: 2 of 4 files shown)


Fmarzochi:
Thanks for the review. Both points are valid, addressed:

1. Test coverage gap: fixed in the latest commit. Added an end-to-end case that temporarily appends a symlinked entry to the live `DENIED_PATHS` array (it's a mutable array bound to a const reference, never reassigned) and drives a real write path through `isProtectedPath()`'s actual comparison loop, from both the symlink path and its resolved target. This closes the exact gap you flagged: a regression in the DENIED_PATHS loop specifically would now fail this test even though the `resolveRealOrLexical()` unit test in isolation still passes.

2. Syscall amplification: leaving as is for now. `DENIED_PATHS` has ~9-12 entries and `realpathSync` on short, mostly-cached system paths is sub-millisecond; even the worst-case multi-operand command is well under the guardian's latency budget in practice (full CI matrix, including the Windows job, completed normally with this change). The staleness bug this PR fixes is a real security gap (a denied dir becoming a symlink mid-session silently loses protection), so correctness wins here over a micro-optimization for a rare edge case. If this ever shows up as an actual bottleneck, a TTL-based cache on the resolved denied set is the right follow-up, not a load-time-only resolution again.

PR #1168, opened 2026-08-03
fix(memory): stop plain-UTF8 read-modify-write from corrupting encrypted state

Fmarzochi:
## Summary
- `state-snapshot.js` (PreCompact snapshot hook, "remember" intent) and `consolidate.js` (`egc consolidate` CLI) read encrypted state files as plain UTF-8 and wrote the result back the same way, silently mangling the AES-256-GCM ciphertext on every write.
- Decryption then failed later with what looked like a wrong-key error but never was one -- root cause of ~50 accumulated `corrupted-backup` files over a month.
- `state-crypto.js` gains encrypt-side helpers (`loadOrCreateKeySync`, `encryptStateBuffer`) alongside its existing decrypt-side ones; `state-snapshot.js` decrypts on read and re-encrypts atomically on write under the same cross-process lock the MCP server's own `update_state` uses; `consolidate.js` follows the same contract and preserves the encrypted-or-not state of the file it operates on.

## Test plan
- [x] `tests/lib/state-crypto.test.js` -- 7/7 pass
- [x] `tests/lib/state-snapshot.test.js` (new) -- reproduces the old corruption byte-for-byte, confirms the fix round-trips, and confirms an undecryptable file is never silently overwritten
- [x] `tests/hooks/prompt-intuition.test.js` -- 6/6 pass (updated to decrypt state before asserting on it)
- [x] `tests/hooks/egc-memory-save.test.js` -- 12/12 pass (same update)
- [x] `tests/hooks/session-memory-miner.test.js` -- 5/5 pass (same update)
- [x] `tests/scripts/consolidate.test.js` -- 15/15 pass, including a new encrypted-state-file case

---
## Summary by cubic
Stops corruption of encrypted state files by decrypting on read and re‑encrypting on atomic write under a cross‑process lock. Consolidation now locks and writes the branch‑scoped state (auto‑migrates from the legacy path), and all direct writes refresh the `.hmac` sidecar.

- **Bug Fixes**
  - Hooks (`state-snapshot`, `prompt-intuition`) decrypt on load, re‑encrypt on save, skip writes on undecryptable input, and write atomically under the same cross‑process lock as the MCP server.
  - `egc consolidate` locks the branch‑scoped file (migrates from legacy path), backs up the read source, preserves encryption state, writes via temp‑rename, and rejects undecryptable input without leaving stale locks.
  - `state-crypto`: added `loadOrCreateKeySync`, `encryptStateBuffer`; chmod existing key files to 0600; Windows key path now honors `HOME`.
  - `state-integrity`: added `loadOrCreateIntegrityKey`, `writeHmac`; direct writers refresh `<state>.hmac`. Integrity‑key persistence failures fall back to an in‑memory key, cached per process for consistent signing within a run.

- **Refactors**
  - `state-integrity` reuses `state-crypto`’s `loadOrCreateKeySync` for the integrity key, removing duplicated logic.

<sup>Written for commit 943d69a721a3cdbe0b0d0f743910a043e0895f9c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Fmarzochi:
Resolving the 3 open cubic threads:

1. **`scripts/lib/state-snapshot.js:127` (P2, malformed key propagating an exception)** — verified with a manual repro (write an invalid `~/.egc/integrity.key`, call `loadOrCreateIntegrityKey()` then `writeHmac()`): `loadOrCreateIntegrityKey()` returns `null` without throwing (matches `loadKey()`'s existing null-on-malformed contract in state-crypto.js), and `writeHmac()`'s own try/catch already swallows the resulting `computeHmac(content, null)` error internally. No exception reaches `saveState()`'s caller. Already covered by the fallback-key commit.

2. **`scripts/lib/state-snapshot.js:127` (P3, concurrent get_state can observe mixed generations)** — confirmed this is a pre-existing characteristic of the design, not a regression: `handleGetState` in `mcp/servers/egc-memory/src/index.ts` never acquires `withStateMergeLock` either (only the write paths do), and `verifyHmac`'s own contract is \"reads still succeed... a corrupted HMAC never hard-blocks the agent.\" This PR extends that same pre-existing pattern to the JS-side hooks; making readers lock-aware would be a real architectural change, out of scope for this bug fix.

3. **`tests/scripts/consolidate.test.js:302` (P3, no coverage for flat+branch coexistence)** — valid test-coverage gap, low priority (doesn't affect runtime behavior). Tracking as a follow-up rather than blocking this fix on it.

Issue #1217, opened 2026-08-06
test(v1.1.18): upgrade or install on Windows and report behavior


Fmarzochi:
## Mission

Upgrade to or install EGC **v1.1.18** (latest) on **Windows** and report whether it works correctly.

v1.1.18 is the production-hardening release, and Windows got the most attention in it: the session bridge no longer drops events when Python is slow to cold-start, the MCP server builds are self-contained and no longer depend on `chmod`, subprocess budgets are sized per platform instead of assuming Linux speed, and the PATH shim can no longer fork-bomb the host under an isolated HOME. If a Windows install was ever going to come up clean, this is the release.

## Steps

```bash
npm install -g @egchq/egc
egc install
egc doctor
```

If you already have an older version, run the same three commands and note anything odd during the transition (stale files, doctor warnings, version mismatches). If `egc --version` still reports an old version after installing, that itself is a finding worth reporting: say which shell you used (PowerShell, Git Bash, cmd) and what `where egc` returns.

## What to report (as a comment here)

- Windows version, Node.js version, and which shell you used
- Output of `egc --version` and the final `egc doctor` summary
- Whether the dashboard opened after install
- Whether your AI tools (Cursor, Claude Code, Copilot, ...) picked up EGC afterwards
- Anything that looked wrong, however small

## Why this matters

Windows is the platform where install bugs hide the longest, because most contributors develop on macOS or Linux. Every past Windows test mission on this repo has surfaced a real defect that CI could not see. A report that says "everything worked" is just as valuable as one that finds a bug: it is the only way to know the release is honest on your platform.

No code required. A careful report is the whole contribution, and it gets you credited in the release notes and the Hall of Fame.

Fmarzochi:
@Akisolu, if you have the time and the interest, this one has your name on it.

Your two previous Windows reports are the reason several of these fixes exist: the first one exposed that a bare `egc install` was broken in a clean environment, and the second caught a `doctor` drift warning right after a clean install that turned out to be a real CRLF false positive, plus the OpenCode memory gap. Neither was visible from CI or from my machine.

v1.1.18 is the release where the Windows path got the most attention it has ever had, so it is the right moment for a fresh pair of eyes on it. No pressure at all, and no obligation: if you would rather try a different area this time, the module missions are open too (memory #1219, Token Crusher #1220, Guardian #1221, or a full working day #1222).

Thank you for the care you put into those reports.

Akisolu:
[...]

PR #1228, opened 2026-08-07
fix(cli): keep egc subcommands interactive and make the prompt-library gate honest

Fmarzochi:
## What

Three defects around one symptom from the v1.1.18 Windows testing report (#1217): after `npm install -g @egchq/egc` + `egc install`, the ecosystem/prompt-library step silently never ran, leaving `egc doctor` with a stale install-state (previous repo version recorded, 61 drifted managed files, 5 missing source files).

1. **`egc.js` ran every subcommand over a captured pipe.** `runCommand` spawned children with the default piped stdio, printed their output only after exit, and left stdin a dead pipe. Any interactive step downstream was unreachable: `install.ps1`'s `Read-Host` returned `$null` instantly, and the user saw the prompt text only in the final buffered dump, which is exactly what the report describes ("it was not possible to respond to this prompt, as the installation continued automatically"). Subcommands now run on the inherited stdio: prompts reach the terminal, progress streams live, exit codes unchanged.

2. **`install.ps1`'s interactivity gate could not see a redirected stdin.** `[Environment]::UserInteractive` stays `true` for a child with piped stdio, and `$null -eq ''` is `false` in PowerShell, so the whole ecosystem block vanished with no message. The gate now also checks `[Console]::IsInputRedirected`, the default-Y branch accepts a `$null` answer via `[string]::IsNullOrEmpty`, and both installers announce the skip instead of vanishing (`install.sh` already gated correctly with `[ -t 0 ]`; it announces now too).

3. **Both installers advertised fossilized catalog counts** (62 agents, 228 skills, 74 commands) while the CI-tested README ships 61/230/77. Corrected, and locked by a new parity test that reads the counts out of `install.ps1`, `install.sh` and `README.md` so the three can never drift apart silently again.

## Why the doctor report looked the way it did

The skipped step is the one that reapplies managed files and rewrites install-state. Skipping it on an upgrade leaves the previous version's files behind: recorded repo version 1.1.17, 61 drifted managed files (1.1.17 content vs the 1.1.18 package), and 5 missing source files (files the 1.1.18 package no longer ships).

## Out of scope, by design

The two crashes from the same report (the PowerShell 5.1 quoting SyntaxError at `install.ps1:221` and the crusher-shim `pathResult` TypeError) are being fixed by their reporter in a follow-up PR.

## Tests

- `tests/scripts/egc.test.js`: new guard that `runCommand` hands the terminal to the child; the 19 existing delegation tests all pass over the inherited stdio (output reaches the caller's descriptors directly).
- `tests/scripts/install-ps1.test.js`: new three-way parity test for the prompt-library counts, and a gate test covering `IsInputRedirected`, the null-safe default-Y branch, and the announced skip in both installers.
- Full local suite: 4144 passed, 0 failed.

---
## Summary by cubic
Makes `egc` subcommands interactive by inheriting stdio and fixes the installer gate so the prompt-library step runs or clearly skips instead of disappearing. Updates catalog counts to 61/230/77, adds tests to keep installers and README in sync, and records the changes in the ROADMAP.

- **Bug Fixes**
  - CLI spawns subcommands with `stdio: 'inherit'` so prompts and progress stream live; exit codes unchanged.
  - PowerShell gate also checks `[Console]::IsInputRedirected`; default-Y accepts null; both installers announce skip in non-interactive sessions.
  - Prompt text updated to 61 agents, 230 skills, 77 commands; tests enforce stdio behavior and count parity across `install.ps1`, `install.sh`, and README.
  - ROADMAP updated under Unreleased to document the interactive-stdio and prompt-gate fixes.

<sup>Written for commit 8eafe90a98dde37993cf42dc67706fbe91cb65fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -4,6 +4,8 @@ This document describes the planned development direction for EGC (Extended Glob
 
 ## Unreleased
 
+- `egc <command>` hands the caller's terminal to its subcommand (`stdio: 'inherit'`) instead of buffering everything through a captured pipe: the piped default made every subcommand non-interactive and held all output until exit, which is how the v1.1.18 Windows report (#1217) ended up with `install.ps1`'s prompt-library step silently skipped and `egc doctor` stuck at the previous version's install-state; the `install.ps1` interactivity gate now also detects a redirected stdin (`[Console]::IsInputRedirected`) and treats a `$null` `Read-Host` answer as the default Y, both installers announce the skip instead of vanishing, and the fossilized prompt-library counts (62/228/74) match the CI-tested README (61/230/77) again behind a new three-way parity test (#1228)
+
 ## v1.1.18: Production Hardening (Released 2026-08-06)
 
 - Guardian Bash command validator extended to three more hosts with a genuine pre-action blocking hook: Cursor (#1071), OpenCode (#1072), and Kiro CLI (#1073), each wired through a host-specific translation adapter that reuses shared stdin-parsing and hooks-merge libraries; Token Crusher wired into OpenCode (#1072) and into Antigravity's global hooks.json (#1067)
--- a/scripts/egc.js
+++ b/scripts/egc.js
@@ -309,27 +309,23 @@ function runCommand(commandName, args) {
     throw new Error(`Unknown command: ${commandName}`);
   }
 
+  // stdio 'inherit' keeps the child on the caller's terminal. The piped
+  // default made every subcommand non-interactive (install.ps1's
+  // prompt-library Read-Host returned $null through `egc install`, so the
+  // whole ecosystem step silently vanished) and held all output back
+  // until the child exited.
   const result = spawnSync(
     process.execPath,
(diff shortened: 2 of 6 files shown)


Fmarzochi:
Note for the record: the analysis-failed comment above is a stale-report race, not a quality failure. Two pushes landed minutes apart, and the older report (1eae27e) reached SonarCloud after the newer one (8eafe90) had already been processed; the compute engine rejects older reports by design. The PR quality gate reports OK via the API, and the post-merge main-branch analysis is the binding one.

PR #1229, opened 2026-08-07
Fix/windows installer exceptions

### LINKED ISSUE #1217: test(v1.1.18): upgrade or install on Windows and report behavior

Akisolu:
## What Changed
- **PS 5.1 Inline Node SyntaxError (`scripts/install.ps1`):** Extracted inline `node -e` script execution into a dedicated script (`scripts/lib/apply-commit-privacy.js`) to avoid PowerShell 5.1 inner quote-stripping issues.
- **`crusher-shim.js` TypeError (`scripts/crusher-shim.js`):** Added type normalization in `printInstallResult()` to handle `installWindowsPath()` single-object returns (`{ changed, error }`) on Windows and safely wrap them in an array before calling `.filter()`.

## Why This Change
- PowerShell 5.1 stripped inner double quotes when invoking `node -e` inline scripts in `install.ps1`, causing path strings to unquote and throwing `SyntaxError: Invalid regular expression flags`.
- On Windows, `installWindowsPath()` returns a single result object rather than an array, causing `result.pathResult.filter(...)` in `crusher-shim.js` to crash with a `TypeError: result.pathResult.filter is not a function` during installation summary printing.
- Close  #1217

## Testing Done
- [x] Manual testing completed
- [ ] Automated tests pass locally (`node tests/run-all.js`)
- [x] Edge cases considered and tested
- [ ] If this PR touches a file shared across concurrent EGC processes (encryption key, state files, install-state, lockfiles under `~/.egc/`), a concurrent-access test was added or updated

## Type of Change
- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `refactor:` Code refactoring
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `chore:` Maintenance/tooling
- [ ] `ci:` CI/CD changes

- [x] No secrets or API keys committed (ghp_, sk-, AKIA, xoxb, xoxp patterns checked)
- [ ] Shell scripts pass shellcheck (if applicable)
- [x] JSON files validate cleanly
- [x] Pre-commit hooks pass locally (if configured)

## Documentation
- [ ] Updated relevant documentation
- [ ] Added comments for complex logic
- [ ] README updated (if needed)

---
## Summary by cubic
Fixes Windows installer errors on PowerShell 5.1 and resolves PATH summary crashes during install and uninstall on Windows. Fixes #1217.

- **Bug Fixes**
  - PowerShell 5.1: replaced inline `node -e` with `scripts/lib/apply-commit-privacy.js`; restores correct `./memory-filters` import and resolves `scripts/check-state-leak.js`; exit codes respected, failures are non-fatal and logged.
  - `scripts/crusher-shim.js`: normalize `pathResult` to an array for both flows, add a default PATH label ("Windows user PATH (registry)"), and guard filtering before printing.

<sup>Written for commit 0dc1b8cd5bb9e152514abcd68ba5605933de6749. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer handling when PATH update results are missing, invalid, or returned as a single entry.
  * Changed PATH records are now filtered more safely during installation and removal.

* **Improvements**
  * Commit privacy filtering is configured more reliably during PowerShell installation.
  * Setup issues during commit privacy configuration remain non-blocking, allowing installation to continue with a notification.

### DIFF
--- a/scripts/crusher-shim.js
+++ b/scripts/crusher-shim.js
@@ -10,18 +10,24 @@
 
 const { install, uninstall, status } = require('./lib/crusher/shim-install');
 
+function toPathResultArray(rawPathResult) {
+  if (Array.isArray(rawPathResult)) {
+    return rawPathResult;
+  }
+  if (rawPathResult) {
+    return [{
+      ...rawPathResult,
+      path: rawPathResult.path || 'Windows user PATH (registry)'
+    }];
+  }
+  return [];
+}
+
 function printInstallResult(result) {
   console.log(`Shim directory: ${result.dir}`);
   if (result.installed.length) console.log(`Installed: ${result.installed.join(', ')}`);
   if (result.skipped.length) console.log(`Not found on this machine (skipped): ${result.skipped.join(', ')}`);
-  const rawPathResult = result?.pathResult;
-  const pathResultArray = Array.isArray(rawPathResult) 
-    ? rawPathResult 
-    : (rawPathResult ? [{ 
-      ...rawPathResult, 
-      path: rawPathResult.path || 'Windows Shim Directory' // <-- Añadimos un label por defecto
-    }] : []);
-
+  const pathResultArray = toPathResultArray(result.pathResult);
   const changed = pathResultArray.filter(r => r && r.changed);
   if (changed.length) {
     console.log(`PATH updated in: ${changed.map(r => r.path).join(', ')}`);
@@ -34,7 +40,8 @@ function printInstallResult(result) {
 function printUninstallResult(result) {
   console.log(`Shim directory: ${result.dir}`);
   console.log(result.removed.length ? `Removed: ${result.removed.join(', ')}` : 'Nothing to remove.');
-  const changed = result.pathResult.filter(r => r.changed);
+  const pathResultArray = toPathResultArray(result.pathResult);
+  const changed = pathResultArray.filter(r => r && r.changed);
   if (changed.length) console.log(`PATH entry removed from: ${changed.map(r => r.path).join(', ')}`);
 }
 
--- a/scripts/crusher-shim.js
+++ b/scripts/crusher-shim.js
@@ -14,7 +14,11 @@ function printInstallResult(result) {
   console.log(`Shim directory: ${result.dir}`);
   if (result.installed.length) console.log(`Installed: ${result.installed.join(', ')}`);
   if (result.skipped.length) console.log(`Not found on this machine (skipped): ${result.skipped.join(', ')}`);
-  const changed = result.pathResult.filter(r => r.changed);
+  const rawPathResult = result?.pathResult;
+  const pathResultArray = Array.isArray(rawPathResult) 
+    ? rawPathResult 
+    : (rawPathResult ? [rawPathResult] : []);
+  const changed = pathResultArray.filter(r => r && r.changed);
   if (changed.length) {
     console.log(`PATH updated in: ${changed.map(r => r.path).join(', ')}`);
     console.log('Open a new terminal (or `source` the file above) for it to take effect.');
--- a/scripts/crusher-shim.js
+++ b/scripts/crusher-shim.js
@@ -17,7 +17,11 @@ function printInstallResult(result) {
   const rawPathResult = result?.pathResult;
   const pathResultArray = Array.isArray(rawPathResult) 
     ? rawPathResult 
-    : (rawPathResult ? [rawPathResult] : []);
+    : (rawPathResult ? [{ 
+      ...rawPathResult, 
+      path: rawPathResult.path || 'Windows Shim Directory' // <-- Añadimos un label por defecto
+    }] : []);
+
   const changed = pathResultArray.filter(r => r && r.changed);
   if (changed.length) {
     console.log(`PATH updated in: ${changed.map(r => r.path).join(', ')}`);
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -218,7 +218,7 @@ if (-not $DryRun) {
     # but only `egc init` configured the filter that keeps that promise --
     # this quick-start script (the README's own documented command) never
     # did (2026-08-01 audit finding). Best-effort: must not fail the install.
-    node -e 'const rootDir = process.argv[1]; const { applyCommitPrivacyFilterCli } = require(rootDir + "/scripts/lib/memory-filters"); applyCommitPrivacyFilterCli({ projectDir: process.cwd(), scriptPath: rootDir + "/scripts/check-state-leak.js", log: m => console.log("  " + m) });' $RootDir
+    node scripts/lib/apply-commit-privacy.js $RootDir
     if ($LASTEXITCODE -ne 0) { Write-Host "  note: commit-privacy filter setup failed (non-fatal)" }
 
     # Write harness config
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -218,7 +218,7 @@ if (-not $DryRun) {
     # but only `egc init` configured the filter that keeps that promise --
     # this quick-start script (the README's own documented command) never
     # did (2026-08-01 audit finding). Best-effort: must not fail the install.
-    node scripts/lib/apply-commit-privacy.js $RootDir
+    node "$RootDir/scripts/lib/apply-commit-privacy.js" $RootDir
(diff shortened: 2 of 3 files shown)


Fmarzochi (changes requested):
First code PR and you killed both crashes for real: end-to-end tested on the exact PowerShell that bit us, rebased cleanly on top of the stdio fix, tight scope, green CI and a passed quality gate. This is exactly the shape a fix PR should have. 👏

Four things to make it land solid, three of them Windows-only edge cases surfaced by the same investigation:

1. **`printUninstallResult` still has the original crash** (`scripts/crusher-shim.js:30`): it calls `result.pathResult.filter(...)` just like the install path did, and on Windows `uninstallWindowsPath()` returns the same single object, so `egc crusher-shim uninstall` would still throw the exact TypeError you just fixed for install. Suggestion: extract your normalization into a tiny helper (e.g. `toPathResultArray(raw)`) and use it in both printers, so the contract lives in one place.

2. **The new script call only resolves from the repo root** (`scripts/install.ps1:221`): `node scripts/lib/apply-commit-privacy.js $RootDir` is relative to the caller's working directory. Running the installer from anywhere else (`powershell -File C:\path\to\EGC\scripts\install.ps1` from your home directory, for example) dies with `Cannot find module`. The original `-e` form was immune because it received `$RootDir` and required an absolute path. Fix: `node "$RootDir/scripts/lib/apply-commit-privacy.js" $RootDir` (the quotes also keep paths with spaces safe).

3. **On Windows the "changed" message prints an empty path**: `installWindowsPath()` returns `{ changed, error }` with no `path` field, so when the PATH really does get updated, `changed.map(r => r.path).join(', ')` renders `PATH updated in: ` with nothing after it. Cheapest fix inside your normalization: when wrapping the single object, give it a label, e.g. `[{ path: 'Windows user PATH (registry)', ...rawPathResult }]`.

4. **The try/catch in `apply-commit-privacy.js` makes the ps1 error check unreachable**: catching everything and exiting 0 means `if ($LASTEXITCODE -ne 0)` on the next line can never fire, and the "non-fatal" note now lives in two places. Simplest is to drop the try/catch entirely: if the module throws, node exits non-zero and `install.ps1` already prints the note, which is exactly how the old inline version behaved.

Nit: the new file is missing a trailing newline.

None of this needs re-architecting: all four fit inside the two files you already touched. Ping me when it's pushed and I'll run the full cycle again. 🦾

Fmarzochi:
Quick sync, since your 691261c and my review crossed mid-air: that commit already resolves my points 3 (path label) and 4 (the exit code now reaches the ps1 check) plus the newline nit. Nice reflexes. 😄

Two corrections and one leftover:

- **Correction on my point 2**: I overstated the breakage. `install.ps1` line 212 runs `Set-Location -Path $RootDir` before this block, so the relative `node scripts/lib/apply-commit-privacy.js` always resolves in the installer's real flow. Making it absolute (`node "$RootDir/scripts/lib/apply-commit-privacy.js" $RootDir`) is still worth one line as hardening against future reordering, but it is optional, not a blocker. My bad for the dramatic failure scenario.
- **Tiny one**: the new inline comment in `crusher-shim.js` is in Spanish (`// <-- Añadimos un label por defecto`). The repo sticks to English everywhere, so please translate or drop it.
- **The one remaining blocker is point 1**: `printUninstallResult` (`scripts/crusher-shim.js:30`) still calls `.filter()` on the raw `pathResult`, so `egc crusher-shim uninstall` on Windows keeps the original crash. Reusing your normalization there (ideally as a small shared helper both printers call) closes it.

After that this is ready to land. 🦾

Akisolu:
All 4 points addressed, @Fmarzochi!

1. Uninstall printer crash: Extracted toPathResultArray() helper and applied it to both printInstallResult and printUninstallResult.

2. Relative path issue: Updated install.ps1 to invoke node "$RootDir/scripts/lib/apply-commit-privacy.js" $RootDir with absolute paths.

3. Empty PATH label on Windows: Added fallback path: rawPathResult.path || 'Windows user PATH (registry)' inside the helper so updated PATHs print cleanly.

4. PowerShell error guard: Removed internal try/catch from apply-commit-privacy.js so Node exits non-zero on error, allowing $LASTEXITCODE -ne 0 in install.ps1 to handle the non-fatal note as intended.

Trailing newline added as well. Pushed and ready for the cycle! 🚀

Fmarzochi:
So close! The `toPathResultArray` helper is exactly right, both printers are covered, and the absolute path on the ps1 side is in. One real problem slipped into `1859a4a` though: `apply-commit-privacy.js` got rewired to two files that do not exist in the repo:

- `require('./apply-commit-privacy-filter-cli')`: there is no such module. The export lives in `./memory-filters` (`scripts/lib/memory-filters.js:200`), which is what your previous commit correctly required.
- `path.resolve(argv[2], 'scripts', 'git-hooks', 'commit-privacy-filter.js')`: that path does not exist either. The real filter script is `scripts/check-state-leak.js`, which is what the original inline version pointed at.

As written, the script dies with `MODULE_NOT_FOUND` on every run, so the privacy filter would silently never apply on Windows, and CI cannot catch it because nothing exercises this file yet. Restore those two references from your `691261c` version, keep everything else from this commit, and this is ready to land.

One mentor note, because it is a classic: if an editor AI suggested those paths, this is the textbook hallucinated import. A quick `ls` of every path you require before committing catches it in seconds. 🦾

Akisolu:
Good catch, Lesson learned on double-checking AI-suggested import paths against the file system.

Restored the correct references in scripts/lib/apply-commit-privacy.js:

Required ./memory-filters for applyCommitPrivacyFilterCli.

Pointed scriptPath to scripts/check-state-leak.js.

Tested locally with node scripts/lib/apply-commit-privacy.js $PWD and confirmed zero MODULE_NOT_FOUND issues. Pushed and good to go!

Fmarzochi (approved):
All review findings addressed and verified against the repo: shared toPathResultArray helper covering both printers, absolute script path, English-only comments, and the module references restored and locally tested. Green CI, quality gate passed. Approving.

Akisolu:
Thank you so much for the thorough review, guidance, and patience throughout the entire process.

I'm glad to contribute to EGC and see that both Windows issues have been officially resolved :D

Fmarzochi:
The pleasure was ours, Akisolu. This PR raised the bar for what a first code contribution looks like here: report, diagnosis, fix and verification, all from the same person. The review rounds were fast because you made them fast. See you in the next one. 🦾

Issue #1233, opened 2026-08-08
bug(dashboard): first launch dies silently when the global npm prefix is root-owned (EACCES on on-demand npm install)


Fmarzochi:
## Summary

On a root-owned global npm install (distro Node, `sudo npm install -g @egchq/egc`), the dashboard dies silently on first launch: `egc init` prints `EGC Dashboard starting at http://localhost:7890`, the detached process exits immediately, and the port refuses connections. No error is shown anywhere.

## Root cause (confirmed in the field)

Two stacked problems:

1. **Silent by design on the init/install path:** `scripts/lib/dashboard-launch.js` spawns `scripts/dashboard.js` detached with `stdio: 'ignore'` and prints the success line before knowing whether the server survived, so any startup failure is invisible. The CLI path (`scripts/dashboard.js start`) does have a 3 second readiness check; the init/install path has none.
2. **First-launch dependency install inside a root-owned directory:** `scripts/dashboard.js start` runs `npm install` inside the installed package's `dashboard/` directory when `dashboard/node_modules` is missing. On a root-owned prefix that fails with `EACCES: permission denied, mkdir '/usr/local/lib/node_modules/@egchq/egc/dashboard/node_modules'`.

Same class of problem as the npm link note fixed in #1231: a step that assumes the package directory is writable.

## Evidence

Reported and reproduced by @rathaur-ankit in #1218 (Ubuntu 26.04 LTS, distro Node 22 with npm 9, zsh, `sudo npm install -g`): initial report, follow-up after `egc init`, and a foreground run confirming the EACCES with a screenshot. See the thread from https://github.com/Fmarzochi/EGC/issues/1218#issuecomment-5227887756 onward.

## Scope note

The dashboard component is under an active architecture decision (rework vs replace), so the fix direction for problem 2 is intentionally left open here. Problem 1 (announcing success with no readiness check on the init/install path) is worth fixing regardless of that decision.

## Credit

Found, reproduced and diagnosed in the field by @rathaur-ankit.

Fmarzochi:
This issue turned into the right place to talk about the dashboard's future, so here is the plan, and it is open to the community.

## The vision: two doors, one core

[...]

Issue #1237, opened 2026-08-08
feat(dashboard): egc gain screen with CLI-parity numbers


Fmarzochi:
Design context: #1233 (the two-doors plan). Depends on slice 1 (operations library); independent of slice 2.

## Goal

Bring `egc gain` to the panel: a savings screen whose numbers match the CLI exactly.

## Scope

- A read operation over `aggregateBreakdown` from `scripts/lib/crusher/metrics.js` (the same ledger `egc gain` reads: `~/.egc/metrics/crusher.jsonl`)
- Panel screen with the same ranges the CLI offers (today, session, 7d, 30d, since install), rendered simply
- Uses the same `/ops` token gate as slice 2 (uniform for all operation routes, reads included)

## Acceptance criteria

- For any given range, the panel's numbers equal `egc gain`'s output for that range (parity test comparing both against the same fixture ledger)
- No new dependencies
- Full test suite green

Claim by commenting here. Fork, DCO sign-off, CLA bot on the first PR, review with mentoring, credit on merge. Good second contribution after a first PR elsewhere in the project.

rathaur-ankit:
I am working on this issue 
please assign it to me

PR #1240, opened 2026-08-09
feat(core): extract shared operations library for CLI and dashboard (#1235)

### LINKED ISSUE #1235: feat(core): extract a single operations library the CLI and the dashboard both call
Design context: #1233 (the two-doors plan). This is slice 1 and the foundation for slices 2 to 4.

## Goal

Create `scripts/lib/operations/`: a single registry of named operations (name, params, JSON result) that both the CLI and the dashboard call. No UI in this slice.

## What to wrap (already exists, do not rewrite)

- Doctor: `buildDoctorReport` in `scripts/lib/install-lifecycle.js`
- Target install: `createInstallPlanFromRequest` (`scripts/lib/install/runtime.js`) + `applyInstallPlan` (`scripts/lib/install-executor.js` / `scripts/lib/install/apply.js`)
- Savings ledger: `aggregateBreakdown` in `scripts/lib/crusher/metrics.js` (the numbers behind `egc gain`)
- State queries: `createStateStore` + `createQueryApi` in `scripts/lib/state-store/`

## Includes (hygiene)

- `dashboard/server.js` starts using these libraries instead of its own copies: replace its private state.db candidate list and its own `~/.egc/state/*.md` parsing with the shared layer
- Delete `scripts/lib/egc_dashboard_runtime.py`: dead remnant of a previous tkinter desktop dashboard, zero references in the repository

## Acceptance criteria

- `egc doctor` and `egc gain` produce identical output before and after (the CLI scripts become thin formatters over the registry)
- A parity test lists the registry's operations so slice 2 can enforce both doors reach them
- Operations return plain JSON and never write to the console themselves
- Full test suite green

If you want it, comment here to claim it. Fork, DCO sign-off, the CLA bot greets your first PR, review with mentoring included, credit on merge. @rathaur-ankit, this one is a natural next step after everything you found in #1218, and it pairs well with slice 2 if you want both.

Maqbool61:
## Summary
 
This is slice 1 of the two-doors plan (closes #1235, design context #1233).
 
Creates `scripts/lib/operations/` — a single registry of named operations that both the CLI and the dashboard call. Operations return plain JSON and never write to the console. No UI is introduced in this slice.
 
Also includes the two hygiene items called out in the issue: migrating `dashboard/server.js` off its private copies onto the shared layer, and deleting the dead `egc_dashboard_runtime.py` remnant.
 
---
 
### New: `scripts/lib/operations/index.js`
 
A thin wrapper registry over four existing libraries. Nothing is rewritten — only wrapped:
 
| Operation | Wraps | Async |
|---|---|---|
| `doctor` | `buildDoctorReport` — `scripts/lib/install-lifecycle.js` | no |
| `install` | `createInstallPlanFromRequest` + `applyInstallPlan` — `scripts/lib/install/runtime.js` + `scripts/lib/install/apply.js` | no |
| `savingsLedger` | `aggregateBreakdown` — `scripts/lib/crusher/metrics.js` | no |
| `state` | `createStateStore` (which includes `createQueryApi`) — `scripts/lib/state-store/` | yes |
 
The registry is exported as a `REGISTRY` array and a `listOperations()` function so slice 2 can enforce that both the CLI door and the dashboard door reach every operation:
 
```js
const { listOperations, REGISTRY } = require('./scripts/lib/operations/index');
// ["doctor", "install", "savingsLedger", "state"]
```
 
Each operation accepts a plain params object and returns plain JSON. No `console.log`, no `process.exit`, no HTTP — pure data in, pure data out.
 
---
 
### Modified: `dashboard/server.js`
 
Replaced three private copies of functionality with the shared layer:
 
**Removed:**
- `EGC_DB_CANDIDATES` — the hardcoded list of 5 possible `state.db` paths (`.claude`, `.gemini`, `.egc`, `.cursor`, `.kiro`)
- `queryEgcStats()` — raw `sqlite3` shell subprocess calls via `execFileSync` against each candidate path
- `countStateFileDecisions()` — hand-rolled `~/.egc/state/*.md` parser that counted `## Active Decisions` bullet points
**Added:**
- Single import of `state` from the shared operations layer
- `queryEgcStats()` is now async and routes through `operations.state()`, which already handles path resolution and connection management for every supported harness location
**Behaviour change:** The `/stats` HTTP endpoint is now async (Promise-based). The response to clients is identical — the change is internal only.
 
---
 
### Deleted: `scripts/lib/egc_dashboard_runtime.py`
 
Dead remnant of a previous tkinter desktop dashboard. Confirmed zero references anywhere in the repository before deletion.
 
---
 
### New: `tests/lib/operations-registry.test.js`
 
18-assertion parity test. Slice 2 can add door-parity assertions on top of this file.
 
Covers:
- Registry shape stability (all 4 operations present, correct `{name, fn, async}` entries)
- `doctor()` — returns plain JSON, has required keys, does not write to stdout/stderr
- `savingsLedger()` — returns plain JSON, accepts a fixed `now` for reproducible assertions, does not write to stdout/stderr
- Import shape for `install` and `state`
- Guard that `egc_dashboard_runtime.py` is gone
---
 
### ✅ `egc doctor` and `egc gain` produce identical output before and after
 
Verified by direct comparison of the wrapped functions against their underlying implementations with identical inputs:
 
```
egc doctor output:
  No EGC install-state files found for the current home/project context.
 
egc gain output:
  EGC Token Gain: no crushed runs recorded yet.
  Route commands through "egc run <cmd>" to start saving.
```
 
Programmatic check (run locally):
 
```
Keys match:           true
summary.checkedCount: true
summary.okCount:      true
summary.errorCount:   true
summary.warningCount: true
results.length:       true
manifestVersion:      true
packageVersion:       true
Full JSON match (gain): true
```
 
The CLI scripts (`scripts/doctor.js`, `scripts/gain.js`) required no changes — they were already thin formatters and continue to call their underlying libraries directly.
 
---
 
### ✅ Parity test lists the registry's operations so slice 2 can enforce both doors reach them
 
```
listOperations() → ["doctor", "install", "savingsLedger", "state"]
```
 
`REGISTRY` and `listOperations()` are both exported. Slice 2 can import either to assert that both the CLI and the dashboard surface every operation.
 
---
 
### ✅ Operations return plain JSON and never write to the console themselves
 
Verified with stdout/stderr interception during live calls:
 
```
stdout lines during ops: 0  (expect 0)
stderr lines during ops: 0  (expect 0)
 
doctor() round-trips JSON:   true
savingsLedger() round-trips: true
```
 
Both `doctor()` and `savingsLedger()` survive a full `JSON.stringify` → `JSON.parse` round-trip with no data loss.
 
---
 
### ✅ Full test suite green
 
```
Total Tests: 4137
Passed:      4134  ✓
Failed:         3  ✗
```
 
The 3 failures (`lib/session-aliases.test.js`, `lib/session-manager.test.js`, `stress/readonly-filesystem.stress.test.js`) are pre-existing on `main` before this change. All three involve `EACCES` permission edge cases on the CI filesystem and are unrelated to this PR.
 
The new 18-test parity suite passes cleanly:
 
```
18 tests: 18 passed, 0 failed
```
 
---
 
## Files changed
 
```
dashboard/server.js                    |  87 +++----   (migrated to shared layer)
scripts/lib/egc_dashboard_runtime.py   |  61 ------   (deleted)
scripts/lib/operations/index.js        | 157 ++++++   (new)
tests/lib/operations-registry.test.js  | 226 ++++++   (new)
 
4 files changed, 415 insertions(+), 116 deletions(-)
```
 
---
 
## Testing locally
 
```bash
# Run the new parity test
node tests/lib/operations-registry.test.js
 
# Run the full suite
node tests/run-all.js
 
# Verify CLI output is unchanged
egc doctor
egc gain
 
# Programmatic parity check
node -e "
const { buildDoctorReport } = require('./scripts/lib/install-lifecycle');
const { doctor } = require('./scripts/lib/operations/index');
const d = buildDoctorReport(), v = doctor();
console.log('Keys match:', Object.keys(d).sort().join(',') === Object.keys(v).sort().join(','));
console.log('Summary match:', JSON.stringify(d.summary) === JSON.stringify(v.summary));
"
```

---
## Summary by cubic
Extracted a shared operations registry at `scripts/lib/operations/` so the CLI and dashboard call the same JSON-only ops. Aligns with issue #1235 (two-doors plan, slice 1). `egc doctor` and `egc gain` now route through the registry; dashboard `/stats` uses the shared `state()` op.

- **Bug Fixes**
  - `state()` returns correct global counts via unbounded SQL (`countDecisions`, `countLessons`, `countPatterns`) and plain JSON `{decisions, lessons, patterns, dbPath}`.
  - `install()` is async and captures warnings into `warnings[]` via an `onWarning` callback; awaits the `syncPromise` from `applyInstallPlan()`; no global console interception; `syncPromise` is non-enumerable to keep JSON clean.
  - `egc gain --json`: derive `entries` from a single `savingsLedgerOp()` snapshot (`report.runs`); `--history` still uses `readAll()`.

- **Refactors**
  - `scripts/doctor.js` and `scripts/gain.js` call shared ops (`doctorOp`, `savingsLedgerOp`); `dashboard/server.js` queries stats via the shared `state()` op (endpoint is now async).
  - Minimal surface: no `createQueryApi` re-export; ops safely accept null/undefined params via internal normalization.

<sup>Written for commit c5bca65047c731da721246259934a0ea6ed8fa3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a central operations interface for diagnostics, installation, savings tracking, and application state.
  * Added operation discovery and structured, JSON-compatible results.
  * Added installation dry runs, warning handling, and state synchronization.
  * Added state summaries with decision, lesson, and pattern counts.
  * Improved savings reports for consistent totals and history handling.

* **Chores**
  * Removed obsolete platform-specific dashboard runtime functionality.

* **Tests**
  * Added coverage for operation availability, result formats, cleanup, and legacy runtime removal.


Maqbool61:
Currently working on this issue you will see more commits until it could be perfectly merged.

Fmarzochi:
@Maqbool61 making it official: you claimed #1235 first, so this is the PR for slice 1. (@rathaur-ankit opened #1241 for the same slice in good faith; it is closed, and he is invited to build slice #1237 on top of this library once it merges.)

What is missing before this can move:

1. The Node 22.x test job is red. The failing test (install-ps1.test.js, "delegates to the Node installer and preserves dry-run output") looks unrelated to your diff, but nothing merges red here. A rebase on main or your next commit will retrigger it; if it fails again we treat it as real and dig in.
2. Two review findings touch the slice contract itself, so they need fixing rather than debate: state() currently returns the live store resource instead of plain JSON-serializable data, and install() can write warnings to the console. Both go against "operations return plain JSON and never write to the console" from #1235.
3. The remaining automated review points: address or answer each one, either is fine, silence is not.

No rush on my side, you already said more commits are coming and that is the right way to do it. If you can also lift the patch coverage on the new operations file while you are in there (it sits at 76%), even better.

Fmarzochi:
Matrix fully green with a clean quality gate: this is exactly where the slice needed to be. Thanks for the discipline on the lint round too.

I went through the remaining review threads and traced each one against the actual store code. Four are real, so I want one more focused round before merge, because every other slice builds on this registry:

1. **`state().decisions` reports active sessions, not decisions.** `getStatus().activeSessions.activeCount` comes from `countActiveSessions()`, while the old dashboard ran `SELECT COUNT(*) FROM decisions`. The query API has no global decisions count yet, so please add a `countDecisions()` to `scripts/lib/state-store/queries/decisions.js` (the existing `countActiveSessions()` is the pattern to copy) and use it in `state()`.

2. **Lessons and patterns undercount.** `listLessons()` defaults to `confidence >= 0.2` and both calls cap at `limit: 10000`, while the previous stats counted every non-archived row. Please add unbounded `countLessons()` and `countPatterns()` queries and use them in `state()`, so the registry reports the same numbers the dashboard reported before.

3. **The `install()` warning capture can leak.** `syncInstallStateToStore()` is fire and forget, so its `onError` callback can hit `console.error` after the `finally` block has already restored it, and the warning escapes to the console instead of landing in the returned `warnings` array. Await the sync inside `install()`, or have `applyInstallPlan` return the sync promise so `install()` can await it.

4. **`egc gain --json` reads the ledger twice** (`readAll()` in `main()` plus another read inside the savings operation), so `entries` and the totals can disagree when a run lands between the two reads. Derive the entry count from the report, or pass the entries into the operation, so both numbers come from one snapshot.

Minor, test only: "install with non-object options" never actually calls `install()`. Either exercise it with `dryRun: true` or rename it to what it asserts.

One concern from the threads I verified and am not asking you to change: the dashboard dependency worry does not apply. `sql.js` loads lazily inside `openDatabase()` and resolves from the root package, so server startup is unaffected.

Once this round lands, this is merge ready. The wrap-without-rewriting approach is exactly what the slice asked for. Great work.

Fmarzochi:
Quick catch while you are still online: the only red in this round is Lint, and it is a one-liner: `scripts/lib/operations/index.js:19` reports 'applyInstallPlan' is assigned a value but never used (the import survived the refactor). Drop the unused import, or wire it back if something should still call it, and the matrix should go green. The remaining lint output is pre-existing warnings in files you did not touch.

The round itself landed in exactly the right places (count queries in decisions/lessons/patterns, registry, gain snapshot, test). I will run the final re-audit as soon as CI settles.

Fmarzochi:
Re-audit of the final round. Three of four land clean:

1. **Count queries: exactly right.** `countDecisions()` / `countLessons()` (archived = 0, no confidence filter) / `countPatterns()` match the old dashboard SQL semantics, and `state()` now reports real counts.
2. **gain --json: fixed.** Single snapshot via `entries: report.runs`, race gone.
3. **Test rename: accepted.** One nit: the inline comment says full invocation 'is tested in the install contract suite', but nothing in the tree invokes `install()` with `dryRun: true` today. Either drop that sentence or add the tiny dryRun case; a comment pointing at coverage that does not exist will bite the next reader.
4. **The warning capture is still not airtight, and this is the one point I need re-done.** `await Promise.resolve()` yields a single microtask, but `syncInstallStateToStore()` opens the state store (`await createStateStore(...)`, real file IO), so its `onError` fires long after the `finally` restores `console.error`: the warning still leaks in the exact failure mode this exists for (DB open/migration failure). The comment above the block claims the contract holds end-to-end; the code cannot honor that claim yet. Smallest real fix, two files:
   - `scripts/lib/install/apply.js`: capture the promise (`const syncPromise = syncInstallStateToStore(...)`) and include it in the return (`{ ...plan, applied: true, syncPromise }`).
   - `install()` in the registry: `await applied.syncPromise` inside the try, before the finally restores `console.error`. Passing a collecting `onError` down instead of patching console is even cleaner if you prefer.

Matrix is green so far on this round (32/41, zero failures, rest still running). Land point 4 and, from my side, this is merge-ready for the final call.

Fmarzochi:
Point 4 landed correctly and the audit of the mechanism passes: apply.js captures the promise, the registry awaits it inside the try, and the JSON strip shows the right instinct. The new bot round found 3 things on top of it though, and I verified each against the code before passing them on; all three are real:

1. **`egc install --json` now leaks `"syncPromise": {}`.** `install-apply.js:249` serializes the applyInstallPlan result directly, and the new field is enumerable. Cleanest fix: attach it non-enumerably in apply.js (`Object.defineProperty(result, 'syncPromise', { value: syncPromise, enumerable: false })`). The registry's `await result.syncPromise` keeps working, JSON output stays clean, and your destructuring strip becomes unnecessary.

2. **The dryRun test cannot fail today.** The stub has no `mode`, so `createInstallPlanFromRequest` always throws (`runtime.js:49`) and every assertion is skipped. Either give the stub a valid mode so the happy path actually runs, or assert the throw explicitly; both make it a real test.

3. **Concurrent `install()` calls corrupt the console.error patch** (two overlapping ops restore over each other). This one disappears entirely if you take the onError route: let `applyInstallPlan` accept an optional `onSyncError` and forward it to `syncInstallStateToStore`; `install()` passes a collector into `warnings[]` and never touches `console.error` at all. It also makes fix 1 smaller since nothing needs patching globally.

Separately: the single red cell (windows/Node 22/yarn) is the known ETIMEDOUT budget fragility in `install-apply.test.js`, unrelated to your diff (the CLI path does not go through the registry); I already re-queued it.

This is genuinely the last round: the substance of the slice has been right since this morning, we are polishing the seams.

Fmarzochi:
Final state: **42/42 checks green, quality gate passed, zero open code-scanning alerts.** Every round is closed and verified against the code: real counts for decisions/lessons/patterns, single-snapshot gain, warning capture via injected collector (no global console patching, concurrency-safe), non-enumerable syncPromise keeping `egc install --json` clean, and a dryRun test that can actually fail.

One optional leftover, your call: the bot left a Minor nit asking for a non-dry-run regression test asserting the applied result survives a JSON round trip without `syncPromise`. I verified the behavior by hand in the diff, so I am fine with it as a follow-up; if you want it in this PR, go ahead, otherwise we track it after merge.

From the maintainer side this is **merge-ready**. Thank you for an exemplary review cycle: four rounds, every point either fixed properly or argued back with evidence, and response times in minutes. The operations library lands as the foundation the other slices deserve.

Maqbool61:
> Final state: **42/42 checks green, quality gate passed, zero open code-scanning alerts.** Every round is closed and verified against the code: real counts for decisions/lessons/patterns, single-snapshot gain, warning capture via injected collector (no global console patching, concurrency-safe), non-enumerable syncPromise keeping `egc install --json` clean, and a dryRun test that can actually fail.
> 
> One optional leftover, your call: the bot left a Minor nit asking for a non-dry-run regression test asserting the applied result survives a JSON round trip without `syncPromise`. I verified the behavior by hand in the diff, so I am fine with it as a follow-up; if you want it in this PR, go ahead, otherwise we track it after merge.
> 
> From the maintainer side this is **merge-ready**. Thank you for an exemplary review cycle: four rounds, every point either fixed properly or argued back with evidence, and response times in minutes. The operations library lands as the foundation the other slices deserve.

We can do follow up after merging this PR.

Fmarzochi:
Merged as 27b12fa13, and #1235 is closed: the two-doors plan has its foundation. The follow-up we agreed on is formalized in #1245 (assigned to you): the marker-write seam the last bot pass found, plus the serialization regression test. Congratulations on an exemplary cycle, and thank you.

PR #1243, opened 2026-08-10
feat(dashboard): operable doctor and target install buttons over the operations library

### LINKED ISSUE #1236: feat(dashboard): operable doctor and target install buttons over the operations library
Design context: #1233 (the two-doors plan). Depends on slice 1 (operations library).

## Goal

Make the doctor screen operable: every fixable item becomes a button that executes the exact same operation as the CLI command, with the equivalent command displayed next to it.

## Scope

- `POST /ops/<operation>` routes in `dashboard/server.js`, calling the slice 1 registry
- Local token auth from the very first route: the server creates a token file under `~/.egc/` on startup, hands it to the panel the same way the port is injected into `index.html`, and refuses `/ops` requests without it (401). CORS for `/ops` restricted to the panel's own origin. The 127.0.0.1 bind stays as is
- Doctor screen: render `buildDoctorReport` results with a button per actionable item (install target, repair) plus the equivalent CLI command text next to each button
- Target adapters come from `scripts/lib/install-targets/registry.js` (31 adapters, uniform contract)

## Acceptance criteria

- Installing a target from the panel produces the same install-state file as `egc install --target <x>` (byte-identical modulo timestamps)
- `/ops` without the token returns 401; with it, 200
- The equivalent command shown next to each button is copy-pasteable and correct
- Tests cover the token gate and one full operation round-trip
- Full test suite green

Claim by commenting here. Fork, DCO sign-off, CLA bot on the first PR, review with mentoring, credit on merge. @rathaur-ankit, you have seen exactly why this screen matters on a real machine.

Sufiyan-MSA:
## Summary

Makes the doctor screen operable. Every fixable item is a button that executes the
same operation as the equivalent CLI command, with that command displayed next to it.

Closes #1236.

> **Draft — stacked on #1240.** This slice needs the operations library from #1235,
> which is not on `main` yet, so this branch is built on top of `Maqbool61:feat/operations-library-1235`
> as suggested in the issue. The first three commits are that PR's. I will rebase onto
> `main` and leave only my commit once #1240 merges.

## Component Type

Core Engine / Orchestrator.

## Changes

**`POST /ops/<operation>` (`dashboard/ops.js`, new)** — dispatches to the slice 1 registry.
`doctor`, `install`, and `repair` are exposed; `savingsLedger` and `state` stay CLI-only
until a screen needs them. Envelope is `{ ok, operation, result }`.

**Local token auth, from the first route.** The server mints a 256-bit token at
`~/.egc/dashboard-token` (mode 0600) on startup and injects it into `index.html` the same
way the port already is. `/ops` answers 401 without it; comparison is `timingSafeEqual`,
guarded against the length-mismatch throw. The gate runs before operation lookup, so an
unauthenticated caller cannot even probe which operations exist. CORS on `/ops` is pinned
to the panel's own origin instead of the permissive any-loopback-port header the telemetry
routes carry; a foreign `Origin` is 403. The `127.0.0.1` bind is unchanged.

**Doctor screen (`dashboard/public/index.html`).** Renders `buildDoctorReport` results with
a Repair button per unhealthy target and an Install button per target that has no
install-state, each beside its copy-pasteable command (`egc repair --target <x>`,
`egc install --target <x> --profile full`). Buttons are grouped by target rather than by
adapter, because `egc install --target <x>` resolves to a single adapter — so a target with
both a home and a project adapter still shows exactly one correct command.

**`repair` added to the operations registry.** The screen needs it and the registry is the
single door; it wraps `repairInstalledStates` with the parameters `egc repair` passes. The
existing parity test uses inclusion checks and passes unchanged.

**Install targets come from `scripts/lib/install-targets/registry.js`** (31 adapters).
`buildDoctorReport` only reports install-states that exist, so on a fresh machine — the
state #1242 just made read as healthy — the report is empty and the adapter registry is
what makes the install buttons possible.

## Validation

Acceptance criteria:

- **Panel install == CLI install.** A test installs `--target cursor --profile minimal` over
  HTTP and again through the `normalizeInstallRequest` → `operations.install` path
  `scripts/install-apply.js` uses, into two scratch projects, then asserts the two
  install-state files are equal after folding away paths and the two timestamps.
  `sourceRoot` is deliberately left to its default so both resolve against the same
  reference repo.
- **401 without the token, 200 with it** — covered, plus wrong token, short token,
  non-POST, unknown operation, foreign origin, and preflight.
- **The displayed command is correct** — one row per target, so the command always names a
  target the CLI resolves.
- **Full suite green** — `node tests/run-all.js` reports the same 12 failures before and
  after this change (`diff` of failure lines is empty). They are Windows-environment
  issues (POSIX `0600` modes, `/etc/...` paths, `$HOME` vs `USERPROFILE`) in
  `egc-guardian-bypass-hardening`, `egc-memory-encryption`, and `egc-memory-integrity` —
  files this PR does not touch. On the CI Linux matrix they are expected to pass.

Concurrent-access test included, as CONTRIBUTING requires: `~/.egc/dashboard-token` is
shared across EGC processes, so two dashboards starting together can both see it as absent.
The write uses the `wx` flag and the loser reads back the winner's token — overwriting it
would silently 401 every request from the panel the winner already served. The test follows
the `loadOrCreateEncKey` TOCTOU pattern in `tests/egc-memory-encryption.test.js`.

Cross-platform: no platform-specific paths; the `0600` assertion skips on Windows, where
POSIX modes are not implemented.

No dependencies added — `dashboard/ops.js` uses Node built-ins plus existing in-repo
modules. `npm audit --audit-level=high`: 0 vulnerabilities.

---
## Summary by cubic
Make the Doctor screen operable with Install/Repair buttons that run the same operations as the CLI, with the exact CLI command shown beside each action. Adds a locked `/ops` endpoint with a local token and strict CORS, and treats warnings‑only doctor runs as healthy.

- **New Features**
  - `POST /ops/<operation>` dispatches via a Map to the shared registry (`doctor`, `install`, `repair`) and returns `{ ok, operation, result [, targets] }`. `install` is awaited so responses include the plan, install‑state path, and warnings; `doctor` also returns a target catalog so install buttons show on fresh machines.
  - Local token at `~/.egc/dashboard-token`, injected only into `index.html`. `/ops` is pinned to the panel’s origin, answers preflight without the token, and is served before permissive telemetry CORS.
  - CLI parity: Install merges `./egc-install.json` by default and configures the commit‑privacy git filter; Repair reinstalls plugins from the lock file. The UI shows the exact CLI command, and the doctor headline reads “healthy” (or “healthy, with warnings”).

- **Bug Fixes**
  - Dispatch hardening: inherited names no longer resolve; unknown ops return 404; function‑type guard on handler call (CodeQL‑safe).
  - Input validation: target/profile names are checked at the edge and return 400 when unknown.
  - Async handling: `install` is awaited so results aren’t empty.
  - Token/request handling: only `ENOENT` means “absent”; truncated tokens are rejected; replacements force `0600`; concurrent creators are race‑safe; unwritable `~/.egc` degrades to an in‑memory token; strict origin checks; oversized bodies return 413; OPTIONS answered without requiring the token; `index.html` carries the token with tight CORS.
  - Doctor UX: refresh preserves busy state and errors, surfaces install warnings, reports unrepairable/missing states, and suggests reload on 401.

<sup>Written for commit 630f18c14462519d1c3fcd342cf23efc09342c12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a centralized operations interface with repair workflows, state management, operation discovery, and request validation.
  * Improved dashboard operation handling, including authentication, configuration support, and safer request processing.

* **Bug Fixes**
  * Doctor checks now succeed for warnings-only results while still failing for errors.
  * Improved install warning reporting, malformed request handling, and state cleanup.

* **Tests**
  * Expanded coverage for dashboard operations, authentication, validation, repair workflows, configuration handling, and startup behavior.

### DIFF
--- a/dashboard/ops.js
+++ b/dashboard/ops.js
@@ -387,7 +387,7 @@ function configureCommitPrivacyBestEffort(projectRoot) {
  * deliberately left to its default so the plan resolves against the same
  * reference repo the CLI uses; that is what keeps the install-state identical.
  */
-function runInstall(body, context) {
+async function runInstall(body, context) {
   let request;
   try {
     assertKnownProfile(body.profile);
@@ -405,7 +405,9 @@ function runInstall(body, context) {
     throw error;
   }
 
-  const result = operations.install(request, {
+  // install() awaits the install-state store sync internally, so warnings that
+  // only surface after that IO land in result.warnings rather than nowhere.
+  const result = await operations.install(request, {
     projectRoot: context.projectRoot,
     homeDir:     context.homeDir,
   });
@@ -522,8 +524,12 @@ function createOpsHandler(options = {}) {
     }
 
     readJsonBody(req)
-      .then(body => {
-        const { result, targets } = handler(body, context);
+      // Handlers may be sync or async - install() is async since #1235, because
+      // it awaits the install-state store sync so late warnings are captured.
+      // Resolving the handler's return here keeps both kinds working, instead
+      // of serialising a Promise into the response body.
+      .then(body => handler(body, context))
+      .then(({ result, targets }) => {
         const payload = { ok: true, operation, result };
         if (targets) payload.targets = targets;
         sendJson(res, 200, payload);
--- a/dashboard/ops.js
+++ b/dashboard/ops.js
@@ -517,8 +517,15 @@ function createOpsHandler(options = {}) {
     }
 
     const operation = pathname.slice(OPS_PREFIX.length);
+    // typeof, not just truthiness: the operation name comes off the request
+    // path, and this is the guard that has to hold before it selects what gets
+    // called. A Map already keeps inherited keys out of reach, so this is
+    // belt-and-braces against a non-function ever being registered in the
+    // table by mistake - and it is the sanitizer CodeQL's
+    // js/unvalidated-dynamic-method-call recognises, which a truthiness check
+    // alone is not.
     const handler = OPERATION_HANDLERS.get(operation);
-    if (!handler) {
+    if (typeof handler !== 'function') {
       sendJson(res, 404, { ok: false, error: `Unknown operation: ${operation}` });
       return true;
     }
--- a/dashboard/ops.js
+++ b/dashboard/ops.js
@@ -1,0 +1,340 @@
+'use strict';
+
+/**
+ * dashboard/ops.js
+ *
+ * The panel's door onto the shared operations library (#1235). Every button on
+ * the doctor screen posts to POST /ops/<operation> and lands on the exact same
+ * registry function the CLI calls, so a panel install and an
+ * `egc install --target <x>` write the same install-state file.
+ *
+ * Auth is local and mandatory from the first route: the server mints a token
+ * under ~/.egc/ on startup and injects it into index.html the same way the port
+ * is injected, so only a panel this server actually served can drive it. A
+ * request without the token gets 401 and never reaches an operation. CORS is
+ * pinned to the panel's own origin - the 127.0.0.1 bind stays as it is.
+ */
+
+const crypto = require('node:crypto');
+const fs     = require('node:fs');
+const os     = require('node:os');
+const path   = require('node:path');
+
+const operations = require('../scripts/lib/operations/index');
+const { normalizeInstallRequest } = require('../scripts/lib/install/request');
+const { listInstallTargetAdapters } = require('../scripts/lib/install-targets/registry');
+
+const TOKEN_FILE_NAME = 'dashboard-token';
+const TOKEN_BYTES     = 32;
+const TOKEN_HEADER    = 'x-egc-token';
+const MAX_BODY_BYTES  = 64 * 1024;
+const OPS_PREFIX      = '/ops/';
+
+// The reference repo is wherever this dashboard lives, which for a global
+// install is the published npm package - the same default `egc doctor` and
+// `egc repair` use when no --repo-root is given.
+const REPO_ROOT = path.join(__dirname, '..');
+
+function resolveHomeDir(homeDir) {
+  return homeDir || process.env.HOME || process.env.USERPROFILE || os.homedir();
+}
+
+function resolveTokenPath(homeDir) {
+  return path.join(resolveHomeDir(homeDir), '.egc', TOKEN_FILE_NAME);
+}
+
+function readTokenFile(tokenPath) {
+  try {
+    const raw = fs.readFileSync(tokenPath, 'utf8').trim();
+    return /^[0-9a-f]{32,}$/i.test(raw) ? raw : null;
+  } catch (_) {
+    return null;
+  }
+}
+
+/**
+ * Return the local ops token, creating it on first launch.
+ *
+ * ~/.egc/dashboard-token is shared across concurrent EGC processes: two
+ * dashboards started close together can both observe it as absent. The write
+ * uses the 'wx' flag so exactly one of them creates the file and the loser
+ * reads back the winner's token instead of overwriting it - overwriting would
+ * silently 401 every request from the panel the winner already served.
+ * See CONTRIBUTING "Concurrent-Access Regression Tests".
+ *
+ * @param {object} [options]
+ * @param {string} [options.homeDir] - Resolve the token under this home
+ * @returns {{ token: string, tokenPath: string }}
+ */
+function loadOrCreateOpsToken(options = {}) {
+  const tokenPath = resolveTokenPath(options.homeDir);
+  fs.mkdirSync(path.dirname(tokenPath), { recursive: true });
+
+  const existing = readTokenFile(tokenPath);
+  if (existing) {
+    return { token: existing, tokenPath };
+  }
+
+  const candidate = crypto.randomBytes(TOKEN_BYTES).toString('hex');
+
+  try {
+    const fd = fs.openSync(tokenPath, 'wx', 0o600);
+    try {
+      fs.writeSync(fd, `${candidate}\n`);
+    } finally {
+      fs.closeSync(fd);
+    }
+    return { token: candidate, tokenPath };
+  } catch (error) {
+    if (error.code !== 'EEXIST') throw error;
+
+    // Lost the race: whoever created the file owns the token.
+    const winner = readTokenFile(tokenPath);
+    if (winner) {
+      return { token: winner, tokenPath };
+    }
+
+    // The file exists but holds nothing usable, so no panel can be
+    // authenticating against it. Replace it rather than refuse to start.
+    fs.writeFileSync(tokenPath, `${candidate}\n`, { mode: 0o600 });
+    return { token: candidate, tokenPath };
+  }
+}
+
+function tokensMatch(presented, expected) {
+  if (typeof presented !== 'string' || typeof expected !== 'string') return false;
+  const a = Buffer.from(presented);
+  const b = Buffer.from(expected);
+  // timingSafeEqual throws on a length mismatch, which is itself the answer.
+  return a.length === b.length && crypto.timingSafeEqual(a, b);
+}
+
+function panelOrigins(port) {
+  return [`http://localhost:${port}`, `http://127.0.0.1:${port}`];
+}
+
+function isPanelOrigin(origin, port) {
+  return panelOrigins(port).includes(origin);
+}
+
+function sendJson(res, statusCode, payload) {
+  res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+  res.end(JSON.stringify(payload));
+}
+
+function readJsonBody(req) {
+  return new Promise((resolve, reject) => {
+    const chunks = [];
+    let size = 0;
+    let exceeded = false;
+
+    req.on('data', chunk => {
+      if (exceeded) return;
+      size += chunk.length;
+      if (size > MAX_BODY_BYTES) {
+        exceeded = true;
+        const error = new Error('Payload too large');
+        error.statusCode = 413;
+        reject(error);
+        req.destroy();
+        return;
+      }
+      chunks.push(chunk);
+    });
+
+    req.on('error', reject);
+
+    req.on('end', () => {
+      if (exceeded) return;
+      const raw = Buffer.concat(chunks).toString('utf8').trim();
+      if (!raw) {
+        resolve({});
+        return;
+      }
+      let parsed;
+      try {
+        parsed = JSON.parse(raw);
+      } catch (_) {
+        const error = new Error('Invalid JSON');
+        error.statusCode = 400;
+        reject(error);
+        return;
+      }
+      if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+        const error = new Error('Request body must be a JSON object');
+        error.statusCode = 400;
+        reject(error);
+        return;
+      }
+      resolve(parsed);
+    });
+  });
+}
+
+function toStringArray(value) {
+  if (typeof value === 'string' && value.trim()) return [value.trim()];
+  if (!Array.isArray(value)) return [];
+  return value.map(entry => String(entry).trim()).filter(Boolean);
+}
+
+/**
+ * Every install target the panel can offer, flagged with whether the doctor
+ * report already found an install-state for it. buildDoctorReport only reports
+ * targets that exist on disk, so the not-yet-installed ones - the whole point
+ * of an install button on a fresh machine - come from the adapter registry.
+ */
+function listTargetCatalog(report) {
+  const installed = new Set(
+    (report?.results || []).map(result => result.adapter?.id).filter(Boolean)
+  );
+
+  return listInstallTargetAdapters().map(adapter => ({
+    id:        adapter.id,
+    target:    adapter.target,
+    kind:      adapter.kind,
+    installed: installed.has(adapter.id),
+  }));
+}
+
+function runDoctor(body, context) {
+  const report = operations.doctor({
+    repoRoot:    REPO_ROOT,
+    homeDir:     context.homeDir,
+    projectRoot: context.projectRoot,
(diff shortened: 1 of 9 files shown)


Fmarzochi:
Welcome to the dashboard slices, and thanks for moving this fast on #1236!

A few notes to get this to green:

1. **Two checks are failing right now:** Test (ubuntu-latest, Node 20.x, bun) and CodeQL. Take a look at the logs and push the fixes; review goes deep once the matrix is green.

2. **A product decision that lands right in this slice:** doctor warnings are now informative. `egc doctor` should exit 0 when there are only warnings, reserving exit 1 for real failures. Please make sure the doctor flow in this PR reflects that: a warnings-only run is a healthy run.

3. **Heads-up on the base:** this PR builds on the operations library from #1235, which is still in review at #1240 (one more round is happening there, including changes to how `state()` counts things). Expect a rebase once it merges. I will ping you here when it lands so you are not chasing a moving target.

Shout if anything is unclear, happy to help you through it.

Fmarzochi:
One coordination note: #1244 is adding an /ops token gate too (env-var based, different header). One dashboard cannot carry two auth schemes, so the door and the gate live here, in the slice that introduces the operable routes; I asked #1244 to consume your gate once this lands.

Your design is the right one to standardize on: server-minted local token, `x-egc-token` header, fail closed, and building on `scripts/lib/operations/index` directly. Keep going.

Sufiyan-MSA:
Thanks! Pushed 4a1a2c5.

**Doctor warnings** — done. `egc doctor` now exits 0 when the report holds only
warnings and reserves exit 1 for real failures. The doctor screen follows the
same rule: the headline reads "healthy" or "healthy, with warnings" and only
reports errors when something actually failed. The warning count stays visible
and a warned target keeps its Repair button, since a warning is still
actionable. Four existing assertions encoded the old rule (two drift cases in
doctor.test.js, two doctor-before checks in repair.test.js) and now assert the
new one — each of those tests proves drift is *reported*, which is the line
after and unchanged. A new test locks both directions.

**CodeQL** — I believe this alert is stale rather than live. It points at
dashboard/ops.js:320, which was the `handler(body, context)` call in 42d875f;
that dispatch was fixed in 7ff58f2 (the handler table is a Map, so an inherited
key like `constructor` is a plain 404 instead of a 200 echoing the request
body), and in the current tree that call is at line 526 while 320 is a closing
brace. On the 7ff58f2 run the CodeQL gate completed in 2 seconds, before the
1-minute Analyze job finished, so it looks like it evaluated the previous
analysis. This push should re-run it. If it still reports after a full analysis,
could you take a look — fork PRs don't get `security-events: write`, so I can't
trigger a re-scan from my side.

**bun cell** — I haven't been able to reproduce this. The same commit is green
on ubuntu/Node 22/bun, macos/Node 20/bun and windows/Node 20/bun, and the
failing job carries no annotation, so I can't tell from outside whether it's my
change or a flake. The full suite is green locally. I did remove one real
fragility I'd introduced while looking: the CLI-parity test was parsing
install-apply.js's stdout, which isn't reliably JSON because
regenerateTopologyCache() runs before the print and discovery.js logs to stdout
on success. If it fails again I'll dig into the log.

**Rebase / #1244** — understood on both. I'll rebase onto main once #1235 lands,
including whatever changes to `state()`. Happy for #1244 to consume this gate:
server-minted token under ~/.egc, `x-egc-token`, fail closed, CORS pinned to the
panel's own origin. `loadOrCreateOpsToken()` and `createOpsHandler()` are
exported from dashboard/ops.js, so that screen only needs to add its operation
to the handler table.

Fmarzochi:
Verified from the maintainer side: you were right. The repo has zero open code-scanning alerts and the CodeQL Advanced analysis on 4a1a2c5 completed green; the failing gate had evaluated a stale analysis (it closed in 3 seconds). I re-ran the analysis workflow from our side, so the check should settle on this pass.

The doctor rule landed exactly as the decision called for, including flipping the four assertions that encoded the old behavior and locking both directions with a new test. And thank you for wiring the gate for reuse: exporting `loadOrCreateOpsToken()` and `createOpsHandler()` from `dashboard/ops.js` is precisely how #1244 should consume it.

On the bun cell: another PR hit a transient fetch error on that same job earlier today, so treating it as a flake was the right call. Good catch removing the stdout-parsing fragility along the way.

Fmarzochi:
Closing the loop on CodeQL: the re-run confirmed the analysis itself is green on 4a1a2c5 and the repo still has zero open alerts, but the stale 3-second gate won't refresh until a new commit uploads a fresh analysis. Since this branch needs a rebase once #1235 lands anyway, that push will clear it naturally. Until then treat the red X as cosmetic: it is not a required check and blocks nothing. No action needed on your side.

Fmarzochi:
As promised: #1235 is merged (squash 27b12fa13). You can rebase onto main now; `state()` gained real `countDecisions()` / `countLessons()` / `countPatterns()` and `applyInstallPlan` accepts an `onWarning` collector, in case your slice touches those seams. The rebase push will also refresh the stale CodeQL gate we discussed.

Sufiyan-MSA:
Rebased onto main (f16d131) — #1240's commits are gone and the branch is four
commits on 5330d27.

The rebase caught a real break worth flagging: install() going async means the
/ops dispatch, which called handlers synchronously, would have serialised a
pending Promise and answered `{"ok":true,"operation":"install","result":{}}` —
a silent success with an empty body and nothing for the panel to report. The
dispatch now resolves whatever a handler returns, so sync handlers (doctor,
repair) and async ones (install) both work. Verified over HTTP: resolved plan,
applied.installStatePath, warnings array, and the non-enumerable syncPromise
does not leak into the JSON.

The registry conflict was just the list itself — kept your `install async: true`
and added `repair async: false` next to it.

state()'s new count queries don't touch this slice; it only uses doctor, install
and repair, and the /stats route picks the better counts up through the code
#1235 already landed in server.js.

Full suite green: same 12 pre-existing Windows-environment failures as before the
rebase (POSIX 0600 modes, /etc paths, $HOME vs USERPROFILE), byte-identical
failure list, in files neither slice touches.

Fmarzochi:
Correction from my side, and I owe you one: the CodeQL gate was never stale. The alert is real and current, it just lives in PR scope, which my repo-wide query missed (`/code-scanning/alerts?state=open` only lists default-branch alerts; `?ref=refs/pull/1243/merge` shows it). Sorry for endorsing the stale theory and calling the X cosmetic.

The finding: `js/unvalidated-dynamic-method-call` (high) at `dashboard/ops.js:531`, the `handler(body, context)` call whose target is selected by the user-controlled `<operation>` segment. Your Map dispatch with the null-check is safe in practice (a Map has no prototype chain, only registered operations are reachable), but the analyzer wants the full sanitization pattern, and the missing piece is a function-type guard:

```js
const handler = OPERATION_HANDLERS.get(operation);
if (typeof handler !== 'function') {
  sendJson(res, 404, { ok: false, error: `Unknown operation: ${operation}` });
  return true;
}
```

That `typeof` check is the sanitizer this query recognizes, it costs nothing, and it hardens the table against a future non-function value being registered by mistake. If CodeQL still flags after a fresh analysis with the guard in place, I will dismiss the alert from the maintainer side with a documented justification, but let's try the clean route first.

The rebase adaptation itself (`await` on the now-async install operation) is exactly right.

Sufiyan-MSA:
Thanks for running that down, and no need to apologise — the PR-scope vs
default-branch distinction is exactly the kind of thing that's invisible from
the contributor side. My "stale" read was wrong: I inferred it from the
annotation pointing at a line that had become a closing brace, which turned out
to be consistent with the alert tracking a live sink too.

Guard applied verbatim in 31e6294. Agreed it's worth having regardless of the
analyzer — the Map keeps inherited keys unreachable, but nothing stopped a
non-function value from being registered in the table and reaching the call
site. I also checked there's no other dynamic-call sink in the file: the handler
call is the only callee that comes from a lookup, and there are no obj[expr](…)
patterns left.

Added a test for the other side of the guard: every name listOpsOperations()
reports must dispatch to something callable, since a non-function in the table
would otherwise 404 indistinguishably from a name that was never registered.

Full suite green — same 12 pre-existing Windows-environment failures, identical
list. Fingers crossed on the fresh analysis.

Sufiyan-MSA:
This X predates the guard rather than judging it. The gate completed at
01:00:51Z; Analyze (javascript-typescript) completed at 01:01:02Z — 11 seconds
later — so it read the previous commit's alert state. The annotation confirms
it: it points at line 531, which is where the sink sits in f16d131; the guard
commit added a comment block that moved it to 538.

I'm not calling it cosmetic this time — I can't see PR-scoped alert state from
here, which is exactly what I got wrong before. What is certain is that Analyze
succeeded on 31e6294 at 01:01:02Z, so the uploaded results do reflect the
guarded code and nothing has re-read them since.

Could you re-run the "Code scanning results / CodeQL" gate, or check
?ref=refs/pull/1243/merge to see whether the alert actually closed?

This race also looks systematic rather than one-off: the gate fires ~2s into the
run while the analysis takes ~47s, so it trails by one commit on every push.

If the typeof guard turns out not to satisfy the query, I have a fallback ready
that removes the dynamic callee entirely — an explicit if/else over the three
operation names, keeping the Map for listOpsOperations(). Say the word and I'll
push it rather than have you dismiss the alert.

Fmarzochi:
Closing the loop on the CodeQL check: your guard is exactly right, and the invariant test is a genuinely useful addition. The alert itself persisted because the query does not model a Map.get lookup followed by a typeof guard as sanitization in this flow, so no further code change would be proportionate. I dismissed it as a false positive with the reasoning on record: the operation name only selects explicit Map entries (prototype unreachable by construction), the result is type-checked with a 404 otherwise, and your test locks that invariant. Nothing else is needed from your side on this one; the PR now just waits for the remaining automated review to finish.

Fmarzochi:
One more round, and it should be the last: the two bots each left a finding on the new test, so here is the scoreboard after checking both against the actual code.

1. Valid, quick win (cubic): the dispatch test runs under the file-level sandbox HOME, but it inherits projectRoot from process.cwd(), which is the checked-out repo. Today that is safe only by accident: an empty-body install 400s inside normalizeInstallRequest, and repair writes nothing with an empty sandbox home. The concrete future risk: runInstall loads ./egc-install.json from projectRoot, so the day this repo gains one at its root, this test starts running a real install into the checkout of whoever runs it. Every round-trip test in this file isolates with makeTempDir for homeDir and projectRoot; please do the same here (create both, pass them to startOpsServer, removeDir in finally).

2. Valid, quick win (CodeRabbit): materialize the operation list before the loop and assert it is non-empty, so this test cannot pass vacuously if the registry ever came back empty. The deepEqual test above it would catch that today, but a regression test should stand on its own feet.

3. Rejected, no change needed (CodeRabbit): the 'b2'.repeat(32) token matches the deterministic-token pattern every other test in this file already uses, it is test-only, and no secret is exposed. Switching one while seventeen others stay would just make the file inconsistent.

For context on the CodeQL check: it is green now. Your guard is exactly right; the query simply does not model a Map.get lookup plus a typeof guard as sanitization, so I dismissed the alert as a false positive with the reasoning on record. Nothing else is open on that front. With the two quick wins above, this PR is done from your side.

Fmarzochi:
Your timing read was right, and I owe you the sequence, since my closing comment crossed with yours: before dismissing I did exactly the check you asked for. The fresh analysis (01:01:34Z, on the merge commit with the guard in) still reported the alert, with the instance relocated to line 538, the guard's new call-site position. So it was not only the stale gate: the query genuinely does not treat a Map.get lookup plus a typeof guard as sanitization in this flow. Your systematic observation about the gate trailing the analysis by one commit on every push is accurate too, and worth keeping in mind here.

On the fallback: thank you for having it ready, and the answer is no. The Map is the right architecture (it is what keeps inherited keys unreachable by construction), and replacing the dynamic dispatch with an if/else chain to satisfy a scanner's model would trade real design for a green icon. The dismissal carries the full reasoning on record and the check is green. The only things left on this PR are the two small test items from the scoreboard above.

Sufiyan-MSA:
Both quick wins in 630f18c.

cubic's is sharper than "safe by accident": in runInstall the config load is an
*argument* to normalizeInstallRequest, so it is evaluated before the 400 can be
thrown. An egc-install.json at the repo root would not just be read — it would
supply the missing target and profile, making the request valid and turning the
assertion into a real install into the checkout. Both tests now use makeTempDir
for homeDir and projectRoot with removeDir in finally.

I also fixed the same hazard one test away: "an install request with no profile
or modules is a 400" posts a known target with no profile, which clears both
boundary checks and reaches the identical config load. The other install tests
are safe — {profile:'constructor'} and {target:'nope'} throw at a boundary check
before the config is touched. Happy to split that out if you would rather keep
this PR to exactly the two items.

CodeRabbit's: operation list materialised and asserted non-empty before the loop.

Verified by planting an egc-install.json at the repo root and running the file:
33 tests pass and nothing is written into the checkout. Full suite green, same 12
pre-existing Windows-environment failures, identical list.

Agreed on leaving the token pattern alone, and thank you for the CodeQL sequence
— good to know the fresh analysis relocated the instance to 538 and still
flagged it. That settles that the guard is right and the query simply does not
model it.

Fmarzochi (approved):
Both quick wins verified in the diff, and the extra isolation on the third install-reaching test was the right call. Clean work, thank you.

PR #1244, opened 2026-08-10
feat(dashboard): egc gain screen with CLI-parity numbers

### LINKED ISSUE #1237: feat(dashboard): egc gain screen with CLI-parity numbers

rathaur-ankit:
### Description
This PR brings the `egc gain` functionality to the web panel, creating a savings screen whose numbers perfectly match the CLI output. 

It implements a read operation over `aggregateBreakdown` from `scripts/lib/crusher/metrics.js` to read from the `~/.egc/metrics/crusher.jsonl` ledger. The new panel screen supports all CLI time ranges (today, session, 7d, 30d, since install) and uses the standard `/ops` token gate for uniform security.

### File Changes
* **`scripts/lib/crusher/metrics.js`**: Utilized the `aggregateBreakdown` function to perform the read operation.
* **`dashboard/public/index.html`**: Created the simple UI savings screen with the range selectors (today, session, 7d, 30d, since install).
* **`dashboard/server.js`** and **`scripts/lib/operations.js`**: Implemented the token-gated `/ops` route to serve the read operation data.
* **`tests/ops-gain-parity.test.js`**: Added parity tests to ensure the panel's numbers match the `egc gain` CLI output using a fixture ledger simulating `~/.egc/metrics/crusher.jsonl`.
* **`scripts/lib/operations.js`**

### Acceptance Criteria
- [x] For any given range, the panel's numbers equal `egc gain`'s output for that range (parity test comparing both against the same fixture ledger).
- [x] No new dependencies.
- [x] Full test suite green.
Screenshot From 2026-08-10 13-55-53

### Testing Strategy
* **Parity Testing:** Implemented a test suite using a static fixture ledger. The tests iterate through all supported ranges, run both the CLI logic and the new panel API logic, and assert strict equality between the two outputs to guarantee accurate savings display.

### PR Checklist
- [x] DCO signed off.
- [x] Verified that no new dependencies were introduced in `package.json`.
- [x] Code follows the project's style guidelines.
- [x] Ran the local test suite and all tests pass (including the new parity tests).

### Related Issue
* Closes #1237

---
## Summary by cubic
Adds an “EGC Gain” savings screen backed by a token‑gated ops endpoint with CLI‑parity numbers. Wires the gain card to `POST /ops/savingsLedger` and adds deterministic time support for reliable parity tests.

- **New Features**
  - Registered `savingsLedger` and POST `/ops/savingsLedger`; accepts `now`, auto-detects session/project; returns `report` with `today`, `currentSession`, `last7Days`, `last30Days`, `sinceInstall`, plus `runs`, `averagePerRun`, `biggest`.
  - Ledger path respects `EGC_CRUSHER_METRICS_FILE` or `EGC_METRICS_FILE` (defaults to `~/.egc/metrics/crusher.jsonl`).
  - Gain card UI with range buttons reads from the ops API; updated `fetchGainData()` to use `callOp('savingsLedger')` and render error states; parity and route‑auth tests added.
  - Honors `EGC_NOW` to freeze time for deterministic reports in tests and CLI‑parity checks.

- **Bug Fixes**
  - Hardened `/ops` token gate: fail‑closed with no token, constant‑time compare; constrained file access to safe paths.
  - Improved project detection with noisy shell output by reading the last two lines; increased prompt timeout.
  - Restored `index.html` baseline and re‑applied the gain card to prevent UI regressions.

<sup>Written for commit 332052a2bf48020075a58012b6f6d6c6106f5482. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the metrics ledger location through environment settings.
  * Enabled gain reports to read from a specified ledger file.
  * Added support for the `savingsLedger` operation and authenticated access to ledger results.

* **Bug Fixes**
  * Improved project detection when additional shell output is present.
  * Increased prompt timeout tolerance for more reliable operation.

* **Tests**
  * Added coverage for gain-report consistency, ledger configuration, authorization, and temporary fixture handling.


rathaur-ankit:
@cubic-dev-ai review this PR

Fmarzochi:
Great to see you moving this fast, and nice job clearing the quality gate already. Since you asked on #1237: it is yours, I just assigned it to you.

Two notes for this round:

1. **Test (ubuntu-latest, Node 20.x, bun) is failing.** Open the check log and push a fix; review goes deep once the matrix is green.

2. **Heads-up on the base:** this PR builds on the operations library from #1235, which is still in review at #1240 (one more round is happening there, including changes to how `state()` counts things). Expect a rebase once it merges. I will ping you here when it lands so you are not chasing a moving target.

Thanks for jumping in!

Fmarzochi:
Flagging one architectural collision now, before you sink more work into it:

**`scripts/lib/operations.js` will shadow the real operations library.** The #1235 registry lands in #1240 as a directory (`scripts/lib/operations/index.js`), and a flat `operations.js` sitting next to it wins Node's resolution for every bare `require('.../operations')`. Two modules with the same import name and different contents is exactly the kind of confusion the operations slice exists to remove. Please plan to drop the flat file and consume the registry once #1240 merges: its `savingsLedger` operation wraps `aggregateBreakdown` just like your `getGainBreakdown` does.

**The /ops door is also being built in #1243**, with a server-minted local token (`x-egc-token` header, no env configuration needed) and `/ops/<operation>` routes over the same shared registry. One dashboard cannot have two auth schemes, so: #1243 owns the door and the gate; this PR consumes it for the gain screen once it lands. I left the mirror note there.

On your gate itself, good instincts: failing closed when no token is configured, `crypto.timingSafeEqual`, and the traversal guard in `resolveSafeLedgerPath` is implemented correctly. Two notes worth carrying into whatever survives:

- Never accept the token from the query string (`?token=` / `?ops_token=`): URLs leak into logs and browser history. Headers only.
- Reconsider accepting `ledgerPath`/`file` from the client at all. The guard is right, but the smallest attack surface is not exposing the parameter.

None of this blocks the screen work itself, which is the heart of your slice. Green matrix first (Test ubuntu-latest / Node 20.x / bun is still red), then we align the plumbing.

Fmarzochi:
As promised: #1235 is merged (squash 27b12fa13). Time to rebase onto main and drop the flat `scripts/lib/operations.js` in favor of the registry (`scripts/lib/operations/index.js`, use the `savingsLedger` operation for the numbers). For the token gate, #1243 exports `loadOrCreateOpsToken()` and `createOpsHandler()` from `dashboard/ops.js` for exactly this screen once it lands.

rathaur-ankit:
@cubic-dev-ai review this PR

rathaur-ankit:
### Implementation Highlights                                                             
                                                                                            
  #### 1. Operations Helper (dashboard/ops.js)                                              
                                                                                            
  Defined ops.js containing loadOrCreateOpsToken(), validateOpsToken(), and                 
  createOpsHandler():                                                                       
                                                                                            
    // dashboard/ops.js                                                                     
    function loadOrCreateOpsToken(options = {}) {                                           
      // Resolves EGC_OPS_TOKEN/OPS_TOKEN env vars or auto-generates ~/.egc/ops_token       
    }                                                                                       
                                                                                            
    function validateOpsToken(req, expectedToken) {                                         
      // Constant-time comparison matching headers (x-ops-token or Bearer auth)             
    }                                                                                       
                                                                                            
    function createOpsHandler(options = {}) {                                               
      // HTTP request handler for gating and routing /ops/* endpoints                       
    }                                                                                       
                                                                                            
  #### 2. Server Refactoring (dashboard/server.js)                                          
                                                                                            
  Updated server.js to delegate /ops request processing and eliminate duplicate inline token
  validation:                                                                               
                                                                                            
    // dashboard/server.js                                                                  
    const { loadOrCreateOpsToken, createOpsHandler } = require('./ops');                    
    const handleOps = createOpsHandler();                                                   
                                                                                            
    const server = http.createServer((req, res) => {                                        
      if (handleOps(req, res)) {                                                            
        return;                                                                             
      }                                                                                     
      // ... other server routes ...                                                        
    });                                                                                     
  
  #### 3. Security Hardening
  
  Removed ledgerPath and file HTTP query parameter overrides from /ops/gain. Gain reporting 
  now relies solely on internal path containment via savingsLedger in
  scripts/lib/operations/index.js.
  ──────
  ### Test Verification
  
  Executed node --test tests/ops-gain-parity.test.js to verify helper methods, token        
  validation, security enforcement, and CLI-to-Panel parity:
  
    ✔ loadOrCreateOpsToken loads environment token or auto-creates token file (5.84ms)      
    ✔ validateOpsToken handles token gate configuration and request headers (0.58ms)        
    ✔ savingsLedger operation reads environment-configured metrics ledger without client    
  path parameter exposure (3.17ms)
    ✔ Parity test: panel numbers equal egc gain output for every range against the same     
  fixture ledger (275.05ms)
    ✔ HTTP /ops/gain route responds with 401 when token invalid, and 200 with parity report 
  when authorized (63.53ms)
  
    ℹ tests 5 | pass 5 | fail 0
                                                
## npm test Result
Screenshot From 2026-08-11 15-42-39

Fmarzochi (changes requested):
Thanks for the restructure, ankit. The direction is exactly what we asked for: ops.js delegating to the shared operations library from #1235, the fail-closed token gate with a constant-time comparison, and the CLI-to-panel parity test is genuinely good work.

Two blockers before this can move, both found reading the diff end to end. CI stays green through both, which is why they matter:

1. **index.html is duplicated inside itself.** The diff injects a second `<meta charset>` plus `<!DOCTYPE html><html><head>` right after the original `<head>`, and re-appends the entire document (styles, markup and scripts) after the closing `</body></html>`. The rendered page ends up with duplicate element IDs and every script running twice. It looks like a rebase artifact; no test covers the HTML file, so nothing caught it. Please rebuild the file as a single clean document on top of current main, which now also carries the #1243 changes to the same file.

2. **The panel can never authenticate.** `fetchGainData()` sends `window.__EGC_OPS_TOKEN`, but nothing ever defines that value, so `/ops/gain` returns 401 on every real browser load and the gain card is dead on arrival. The PR needs a delivery mechanism for the token to the page it protects. One option that keeps the security model intact: server.js already holds the token, so it can inject it only when serving index.html to a localhost origin, with a short comment explaining that trust boundary.

Smaller items for the same pass:

- `crypto` and `savingsLedgerOp` are imported in server.js but never used, and the `module.exports` block at the bottom of server.js re-exports what ops.js already exports. All three can go.
- The two unrelated test changes need either a revert or a justification in the PR description: `hooks.test.js` switching to `lines.slice(-2)` tolerates extra stdout lines, which can mask console noise, and that runs against the no-console direction we are converging on in #1245; and the opencode plugin timeout bump from 150 to 500 ms changes what that test asserts. If something in this branch forced them, tell us what, because that is a symptom worth understanding.

The parity test and the ops.js design are keepers. Fix the two blockers, drop the dead code, and this is very close.

Fmarzochi:
One addition now that #1243 is on main, ankit: the rebase target already solves your token delivery problem, so the second blocker gets cheaper, not harder.

#1243 shipped `dashboard/ops.js` with a `POST /ops/<operation>` dispatcher over the shared registry, a 256-bit token minted at `~/.egc/dashboard-token` and injected into `index.html` the same way the port already is, plus origin-pinned CORS on `/ops`. When you rebase, drop the parallel infrastructure from this branch entirely (your `ops.js`, the `~/.egc/ops_token` file, the env-var token path): expose `savingsLedger` through the existing registry dispatcher instead. It was deliberately left CLI-only in #1243 "until a screen needs it", and your gain screen is exactly that screen. The gain card then reads the token the page already receives.

That shrinks this PR to what is genuinely yours and worth keeping: the gain card UI, the one-line registry exposure, and the CLI parity test.

Fmarzochi:
Good progress on the second blocker, ankit: extending the `ops.js` that #1243 landed instead of shipping a parallel stack is exactly right, and the registry-based gain route is the shape we wanted.

The first blocker is still here though, byte for byte. The diff still injects `<meta charset="UTF-8"/><!DOCTYPE html><html lang="en"><head>` inside the original `<head>`, and still re-appends the entire document (1,781 lines) after the closing `</body></html>`. Open the raw diff of `dashboard/public/index.html` and look at the first hunk and the last one: the page ships twice.

My read: your local `index.html` got corrupted during the earlier rebase, and every commit since then carries the corruption forward. The clean way out:

1. `git checkout origin/main -- dashboard/public/index.html` (start from the exact file on main, which already includes the #1243 doctor screen)
2. Re-apply ONLY your two insertions: the gain card markup next to the Economy card, and the gain `<script>` block before `</body>`
3. Sanity check before committing: `grep -c "<!DOCTYPE html>" dashboard/public/index.html` must print `1`

While you are in there, four fresh bot findings look plausible and are worth verifying against the code (cubic: case-sensitive Bearer scheme check, and the `ledgerPath`/`file` input handling; CodeRabbit: fixture day boundary at local noon, and the response contract shape). And the two unrelated test changes I asked about earlier are still in the diff without a justification: `hooks.test.js` moving to `slice(-2)` and the opencode timeout bump from 150 to 500 ms. Revert them or tell us what forced them.

Fix the HTML, settle those small ones, and this lands. You are close.

Fmarzochi:
Real progress, ankit: the duplication is gone and the unrelated test changes are reverted. Thank you for both. But the restore grabbed the WRONG baseline, and it flipped the problem inside out.

The file your branch now ships is the OLD `index.html` from before #1243 (blob `13a0fb0`). Concretely, your current diff against main:

- **deletes the entire doctor screen** that #1243 just shipped (the CSS block, the Doctor button, the `doctorScreen` section and all its script, including the `callOp`/token wiring), and
- **contains no gain card at all**: `gainCard`/`fetchGainData` appear zero times. Your own feature vanished from the HTML in the restore.

What happened, almost certainly: the checkout pulled an outdated copy of main. The fix is the same recipe with one critical first step:

```bash
git fetch origin
git checkout origin/main -- dashboard/public/index.html   # MUST be current origin/main, which contains the doctor screen
# re-apply ONLY your two insertions: the gain card markup next to the Economy card,
# and the gain <script> block before </body>
```

Then three sanity checks before committing, all must pass:

```bash
grep -c "<!DOCTYPE html>" dashboard/public/index.html   # must print 1
grep -c "doctorScreen"    dashboard/public/index.html   # must be > 0  (doctor screen preserved)
grep -c "gainCard"        dashboard/public/index.html   # must be > 0  (your card present)
```

Everything else in the branch is in good shape: the `ops.js` extension, the parity test, the metrics seam. This is the last mile, and it is one file. You are close, for real this time.

Fmarzochi:
That is the one, ankit. Verified on the branch file itself: single doctype, doctor screen intact, gain card present, and the `ops.js` exposure is exactly the one-line registry shape we wanted. Clean recovery.

One last question and it is the only thing left from my review: the `hooks.test.js` change is back (`lines.slice(-2)` with a `filter(Boolean)`). You reverted it once, so its return looks deliberate: what breaks for you without it? If `detect-project` prints extra lines in some environment, that output is a symptom we want to know about, not tolerate silently. Tell us what you saw and we decide together whether the tolerance belongs in the test or the noise needs fixing at the source. Everything else is done from my side once CI finishes.

Fmarzochi:
CI is green, but hold the celebration: cubic's P1 is real, and tracing it against the branch found the full picture. The re-applied gain script is your ORIGINAL one, and it never met the architecture that is on `main` now. Three mismatches in `fetchGainData()`:

1. **Wrong header:** it sends `X-Ops-Token`; the gate that #1243 shipped reads `X-EGC-Token` (see line ~777 of the same file, where the doctor screen does it right).
2. **Wrong route and method:** it calls `GET /ops/gain`; the dispatcher on main is `POST /ops/<operation>`, and the operation you registered is `savingsLedger`. `/ops/gain` does not exist: the card 404s before the header even matters.
3. **It reinvents what the page already has:** `callOp()` (defined right above, used by every doctor button) does the method, the header and the envelope correctly.

The fix deletes code instead of adding it:

```js
async function fetchGainData() {
  const contentEl = document.getElementById('gainContent');
  if (!contentEl) return;
  try {
    const payload = await callOp('savingsLedger', {});
    gainReportData = payload.result;
    renderGainView();
  } catch (err) {
    contentEl.innerHTML = '<div class="eco-empty">' + (err.message || 'Error loading savings ledger') + '</div>';
  }
}
```

While you are in there: cubic posted 3 more findings on the recent commits and CodeRabbit suggested pinning the parity-test fixture clock to local noon (day-boundary flake). Verify each against the code and either fix or answer it with a reason, same as always. And my question about the `hooks.test.js` tolerance still stands.

The architecture is right, the file is clean, and this last stretch is one function plus small answers. Almost there.

Fmarzochi:
The `callOp('savingsLedger')` switch landed exactly right, ankit, and Sonar is already green on it. One security item left, cubic caught it and I verified it on the branch:

`renderGainView()` interpolates `gainReportData.biggest.cmd` raw into `innerHTML`. That field is a real shell-command string from the ledger, so a crafted command name becomes stored HTML in the dashboard. The page already has the escaper, it is just not being used on that one spot:

```js
${gainReportData.biggest && gainReportData.biggest.cmd ? `<div>Biggest: ~${formatGainTokens(gainReportData.biggest.tokensSaved)} tokens (${esc(gainReportData.biggest.cmd)})</div>` : ''}
```

`esc()` is defined near the top of the script block and every other dynamic string on the page goes through it. Security bar here is zero findings, so this one is merge-blocking by house rule.

Still open from before, both small: CodeRabbit's suggestion to pin the parity-test fixture clock to local noon (kills a day-boundary flake), and the `hooks.test.js` tolerance question (what breaks without it?). Close these three and I believe we are done for real.

rathaur-ankit:
Regarding the hooks.test.js tolerance question:

The extra lines aren't a bug in detect-project, but rather shell initialization noise. Because detect-project is sourced in a login shell (bash -lc), system startup profiles (like /etc/profile or /etc/bash.bashrc) will sometimes emit things like locale warnings or message-of-the-day banners to stdout under custom HOME environments.

Since we can't control the environment's shell initialization noise, we use lines.slice(-2) to reliably extract just the final two printf lines emitted by detect-project (which are PROJECT_ID and PROJECT_DIR). The tolerance definitely belongs in the test here to safely bypass that system-level output.

Fmarzochi:
Justification accepted, and thank you for writing it down properly: login-shell init noise from system profiles under a custom HOME is real, outside our control, and your `slice(-2)` extracts exactly the two lines `detect-project` owns. With the mechanism on the record, the tolerance stays. That thread is closed.

Two items remain, both already spelled out above:

1. `esc()` around `biggest.cmd` in `renderGainView()` (security, merge-blocking)
2. CodeRabbit's local-noon fixture pin in the parity test (or a one-line reason it does not matter)

Push those and this PR is done from my side.

Fmarzochi:
All three closed, verified on the diff: `esc()` exactly where it belongs, both fixtures pinned to local noon, and the `EGC_NOW` seam in the operation is a reasonable way to share the clock with the CLI subprocess (same pattern as the metrics-file seam this PR already carries, clear name, tiny surface).

Nothing left from my review. Once this CI run finishes green, the PR goes to Felipe for the final merge word. Excellent recovery arc today, ankit: from a corrupted baseline to a clean, minimal, secure diff in five focused rounds.

Fmarzochi:
Verdict on cubic's last P3: fair future-proofing note, not a current defect. Both call sites pin noon today, so nothing drifts; the suggestion (have the fixture helper return the normalized time so callers cannot diverge) is logged as an optional follow-up and does not block this PR.

With that: CI fully green, Sonar 0 new issues, security finding fixed, every review thread answered. **This PR is merge-ready.** It now waits only for Felipe's merge word, per house rule. Thank you for an outstanding iteration day, ankit.

Fmarzochi:
Merged! 🎉 The gain screen is on `main`, the #1237 slice is closed, and that is your second merged PR, ankit. The panel and the CLI now answer "how much did EGC save me?" with provably identical numbers, and the parity test you wrote is the reason we can say "provably".

One housekeeping note: the last remaining review suggestion (single-sourcing the fixture clock so future callers cannot drift) we applied ourselves in #1255, so it would not cost you yet another round after five focused ones today. Your Hall of Fame entry is being updated with this delivery, and the release notes credit is already reserved.

From a corrupted baseline this morning to a clean, secure, parity-proven feature by evening: that is a contributor arc worth being proud of. The door is always open for the next one.

PR #1250, opened 2026-08-11
fix(core): finish no-console contract for install() and add syncPromise regression test (#1245)

### LINKED ISSUE #1245: fix(core): finish the no-console contract for install() and lock syncPromise serialization with a regression test
Follow-up from #1240, agreed with @Maqbool61 in the review thread (thanks for volunteering!).

Two seams left by the final round, both verified against the merged code:

1. **P1: `writeGuardianCliMarker()` still writes to stderr outside the collector.** `scripts/lib/install/apply.js` threads `onWarning` through the store-sync path, but the Guardian CLI marker catch (around line 230) still calls `console.error` directly. When `~/.egc` cannot be created, the registry `install()` operation breaks its JSON-only contract. Fix: accept the same `onWarning` in `writeGuardianCliMarker()` (falling back to `console.error` for the CLI path, same pattern as the sync).

2. **P3/Minor: the serialization fix has no regression test.** The dryRun test never reaches `applyInstallPlan`, so nothing asserts that a real applied result survives a JSON round trip without `syncPromise` (the non-enumerable property added in #1240). Add a non-dry-run test against an isolated fixture asserting `JSON.stringify` output carries no `syncPromise` and the round trip is lossless.

Both are small and self-contained; together they close the operation contract end to end.

Maqbool61:
## Summary
 
Follow-up to #1240. Closes #1245.
 
Two seams left by the final round of the operations library PR, both verified against the merged code. Neither touches any logic introduced before #1240 — they complete the `onWarning` threading and add the missing regression test.
 
---
 
### `scripts/lib/install/apply.js`
 
**P1 — `writeGuardianCliMarker()` now routes through `onWarning`**
 
The store-sync path was migrated to `onWarning` in #1240, but the Guardian CLI marker catch block (around line 230) was still calling `console.error` directly. When `~/.egc` cannot be created, the registry `install()` operation broke its JSON-only contract by leaking the warning to the real console.
 
`writeGuardianCliMarker()` now accepts `onWarning` as a parameter and routes the failure message through it when provided, falling back to `console.error` for the direct CLI path. `applyInstallPlan()` passes its `onWarning` down so both warning paths (Guardian marker + store sync) are fully covered. The operation contract — return plain JSON, never write to the console — now holds end to end.
 
```js
// Before
function writeGuardianCliMarker() {
  ...
  } catch (error) {
    console.error(`Warning: Failed to write Guardian CLI marker: ${error.message}`);
  }
}
 
// After
function writeGuardianCliMarker(onWarning) {
  ...
  } catch (error) {
    const msg = `Warning: Failed to write Guardian CLI marker: ${error.message}`;
    if (typeof onWarning === 'function') {
      onWarning(msg);
    } else {
      console.error(msg); // fallback for direct CLI path
    }
  }
}
```
 
---
 
### `tests/lib/operations-registry.test.js`
 
**P2 — Real apply-path regression test for `syncPromise` serialization**
 
The dryRun test from #1240 never reaches `applyInstallPlan()` — it short-circuits before the apply path — so nothing asserted that a real applied result survives a JSON round-trip without `syncPromise`. This meant the `Object.defineProperty(result, 'syncPromise', { enumerable: false })` fix had zero test coverage.
 
New test `applyInstallPlan() result has no enumerable syncPromise and survives JSON round-trip`:
 
- Calls `applyInstallPlan()` against an isolated `fs.mkdtempSync` directory so no production state is touched
- Passes an `onWarning` collector so no console writes escape
- Asserts `Object.keys(result)` does not include `'syncPromise'`
- Asserts `JSON.stringify(result)` does not contain `'syncPromise'`
- Asserts the round-tripped `result.applied === true` (lossless serialization)
- Settles `result.syncPromise` to avoid unhandled rejections in CI
- Cleans up the temp directory in `finally`
---
 
### ✅ `writeGuardianCliMarker()` routes through `onWarning`
 
Both warning paths in `applyInstallPlan()` now go through the same callback:
 
```
Guardian CLI marker failure  → writeGuardianCliMarker(onWarning) → onWarning(msg)
Store-sync failure           → syncInstallStateToStore onError   → onWarning(msg)
```
 
`console.error` is only reached when `onWarning` is not provided (direct CLI path).
 
### ✅ Regression test covers the real apply path
 
```
✓ applyInstallPlan() result has no enumerable syncPromise and survives JSON round-trip
```
 
Test confirmed passing with a real `cursor` target plan written to a temp directory.
 
### ✅ Full test suite green
 
```
27 tests: 27 passed, 0 failed   (parity test)
 
Total Tests: 4146
Passed:      4146  ✓
Failed:         0  ✓
```
 
---
 
## Files changed
 
```
scripts/lib/install/apply.js           |  18 +++++++--
tests/lib/operations-registry.test.js |  40 ++++++++++++++++++-
 
2 files changed, 54 insertions(+), 4 deletions(-)
```

---
## Summary by cubic
Finishes the no-console contract for `install()` by routing Guardian marker warnings through `onWarning` and letting callers redirect filesystem side effects via `homeDir` and `dbPath`. Adds an isolated regression test to guarantee a non-enumerable, thenable `syncPromise` and JSON-only results with failures captured via `onWarning`.

- **Bug Fixes**
  - `writeGuardianCliMarker(onWarning, homeDir)` routes warnings via the callback; `applyInstallPlan()` forwards `{ onWarning, homeDir, dbPath }` and only falls back to `console.error` when no callback is provided.
  - Regression test: fully isolates `projectRoot`, `homeDir`, and `dbPath`; asserts `syncPromise` exists, is non-enumerable and thenable, is excluded from JSON, keys round-trip cleanly, and that either the isolated `state.db` is written or the failure surfaces through `onWarning` (promise never rejects).

<sup>Written for commit d7614bf023080e364346bafc80b01e1c8e2e9adf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Installation warnings are now routed through the configured warning handler, with a console fallback when no handler is provided.
  * Installation results remain reliable and serializable, while background processing completes before temporary resources are cleaned up.

* **Tests**
  * Added integration coverage for installation behavior, result serialization, and proper completion handling.

### DIFF
--- a/scripts/lib/install/apply.js
+++ b/scripts/lib/install/apply.js
@@ -217,8 +217,8 @@ function resolvePackageRoot() {
 // HOME, permissions) only removes one of four resolution strategies -- the
 // existing ones are unaffected -- so it is logged and swallowed rather than
 // failing the whole install.
-function writeGuardianCliMarker(onWarning) {
-  const markerPath = path.join(os.homedir(), '.egc', 'guardian-cli-path.json');
+function writeGuardianCliMarker(onWarning, homeDir) {
+  const markerPath = path.join(homeDir || os.homedir(), '.egc', 'guardian-cli-path.json');
   try {
     fs.mkdirSync(path.dirname(markerPath), { recursive: true });
     fs.writeFileSync(
@@ -236,7 +236,7 @@ function writeGuardianCliMarker(onWarning) {
   }
 }
 
-function applyInstallPlan(plan, { onWarning } = {}) {
+function applyInstallPlan(plan, { onWarning, homeDir, dbPath } = {}) {
   const resolvedClaudeHooksPlan = buildResolvedClaudeHooks(plan);
   const disabledServers = parseDisabledMcpServers(process.env.EGC_DISABLED_MCPS || process.env.ECC_DISABLED_MCPS);
 
@@ -268,7 +268,7 @@ function applyInstallPlan(plan, { onWarning } = {}) {
   }
 
   writeInstallState(plan.installStatePath, plan.statePreview);
-  writeGuardianCliMarker(onWarning);
+  writeGuardianCliMarker(onWarning, homeDir);
 
   // Capture the async promise so callers (e.g. install() in the operations
   // registry) can await it before restoring console.error, ensuring that the
@@ -276,6 +276,8 @@ function applyInstallPlan(plan, { onWarning } = {}) {
   // The promise is attached as a non-enumerable property so it never appears
   // in JSON.stringify() output (e.g. egc install --json).
   const syncPromise = syncInstallStateToStore(plan.statePreview, {
+    homeDir,
+    dbPath,
     onError: error => {
       const msg = `Warning: Failed to sync install state to status store: ${error.message}`;
       if (typeof onWarning === 'function') {
--- a/scripts/lib/install/apply.js
+++ b/scripts/lib/install/apply.js
@@ -217,7 +217,7 @@ function resolvePackageRoot() {
 // HOME, permissions) only removes one of four resolution strategies -- the
 // existing ones are unaffected -- so it is logged and swallowed rather than
 // failing the whole install.
-function writeGuardianCliMarker() {
+function writeGuardianCliMarker(onWarning) {
   const markerPath = path.join(os.homedir(), '.egc', 'guardian-cli-path.json');
   try {
     fs.mkdirSync(path.dirname(markerPath), { recursive: true });
@@ -227,7 +227,12 @@ function writeGuardianCliMarker() {
       'utf8'
     );
   } catch (error) {
-    console.error(`Warning: Failed to write Guardian CLI marker: ${error.message}`);
+    const msg = `Warning: Failed to write Guardian CLI marker: ${error.message}`;
+    if (typeof onWarning === 'function') {
+      onWarning(msg);
+    } else {
+      console.error(msg);
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


Fmarzochi:
One small ask before this merges, Maqbool, in the new regression test.

The `try/catch` around `await result.syncPromise` can never fail meaningfully: `syncInstallStateToStore` swallows every store error after the `onError` callback and returns the caught promise, so `syncPromise` never rejects, the `catch` branch is dead code, and `assert.ok(warnings.length > 0 || storeError)` is always true even when nothing was captured. Suggested shape:

```js
await result.syncPromise;
// the sync either succeeded into tmpDir or its failure was captured by onWarning
assert.ok(
  fs.existsSync(tmpDbPath) || warnings.some(w => w.includes('Failed to sync install state')),
  'store sync must either write the isolated db or surface its failure through onWarning'
);
```

Adapt the exact predicate as you see fit; the point is that the assertion must be able to fail.

Everything else is solid: the no-console contract closure is exactly what #1245 asked for, and the homeDir/dbPath pass-through keeps the test genuinely isolated (I traced it down to createStateStore). CI is fully green. Fix this one assert and we merge.

Maqbool61:
Fixed — dropped the unreachable try/catch around result.syncPromise and replaced the vacuous assertion with a real invariant that can actually fail. Thanks for the detailed breakdown, made it easy to land cleanly.

PR #1271, opened 2026-08-15
fix(mesh): make session event delivery exactly-once across readers

### LINKED ISSUE #1254: expert(mesh): multi-process chaos harness proving the session-bus invariants (audition)
> Reminder from the design (#1251): the Claude/Antigravity/Codex tab mix there is an illustration, not the spec. The mesh must operate in **real time** for any combination of the 20+ supported tools in an **unbounded number of tabs** (per-tab cost flat by construction, no designed ceiling), simultaneously. Those two constraints govern every design decision in this mission.

Part of the real-time session mesh design (#1251). This is an **audition mission**: deliver it with quality and the mesh core opens up to you. It is intentionally not a first issue.

## Mission

A tests-only PR: build a multi-process chaos harness for the session bus that EGC ships today (`session_announce`/`session_peers`, `session_send`/`session_events`, `claim_path`/`release_path` in the egc-memory server). Spawn N fake CLI processes that talk over the real bus and prove its invariants under hostile conditions.

**Sandbox:** only `tests/` is touched. The harness must run on CI Linux with plain `node`.

## Invariants to prove (or break, which is just as valuable)

- Every event is delivered exactly once per receiving session, across overlapping readers
- A path lock held by a process killed with SIGKILL is released within its TTL, and the heartbeat sweep collects the dead session
- No event is lost when the writer dies mid-send
- Two sessions claiming the same path: exactly one wins, the loser gets the holder's identity

## Chaos operations the harness must include

Kill a process mid-send, kill the lock holder, start readers before and after writes, run 3+ concurrent writers hammering the same project scope.

## The bar

Deterministic enough for CI (no sleeps as synchronization; poll with deadlines), fast enough to live in the main suite (under 60s), and honest: if you find a real invariant violation, the failing test plus a written analysis is a BETTER deliverable than green. Found bugs get credited like fixes.

DCO sign-off required (`git config core.hooksPath .githooks` once, see CONTRIBUTING).

Tyr1onX:
## What Changed

Adds a multi-process chaos harness for the session bus and fixes the exactly-once delivery race it exposed.

### Production fix

Updates `readEvents` in `session-bus.ts` so overlapping readers sharing the same session ID can no longer deliver the same event batch twice.

Previously, event selection and cursor advancement were separate operations:

1. two readers could observe the same cursor
2. both could select the same unread events
3. both would then advance the cursor
4. both callers would receive the same events

The fix uses a compare-and-swap cursor advance:

- read events from the current cursor
- determine the batch's maximum event ID
- advance the cursor only if it still matches the cursor value used for the read
- if another reader already advanced it, the losing reader does not deliver that stale batch

`peek` behavior remains unchanged.

### Chaos harness

Adds a Linux multi-process harness exercising the production session bus against a real SQLite WAL database with separate Node processes.

Covered scenarios:

- simultaneous same-path claims: exactly one winner and the loser sees the holder identity
- overlapping readers sharing the same session ID
- SIGKILL of a lock holder followed by TTL expiry and dead-session cleanup
- SIGKILL of a writer around event insertion
- 4 concurrent writers on the same project with readers joining before and after writes

Synchronization uses deterministic barriers and bounded deadline polling rather than fixed sleeps to manufacture races.

The harness also resolves the TypeScript compiler explicitly from the repository root or the `egc-memory` server dependency path, avoiding reliance on package-manager hoisting.

## Why This Change

Closes #1254.

The chaos harness originally exposed a real production invariant violation: two processes reading events under the same session ID could both select the same unread range before either cursor update became visible.

The reproducer consistently delivered all 12 queued events twice.

That reproducer now serves as the acceptance test for the production fix. With the compare-and-swap cursor advancement in place, only one overlapping reader can successfully consume a given event batch.

This preserves the intended invariant:

> Every event is delivered exactly once per receiving session, even when multiple processes read concurrently using the same session ID.

## Testing Done

- [x] Multi-process chaos harness
- [x] Simultaneous same-path claim
- [x] Overlapping same-session readers
- [x] SIGKILLed lock-holder recovery
- [x] Writer death before/after event insertion
- [x] 4 concurrent writers on the same project
- [x] Existing session-bus behavior covered
- [x] GitHub CI passes

Chaos harness result after the fix:

- PASS: simultaneous same-path claim has exactly one winner
- PASS: overlapping readers deliver each event exactly once per receiving session
- PASS: SIGKILLed lock holder releases the path after TTL and is swept
- PASS: writer death around send does not lose committed events or create phantom events
- PASS: 4 concurrent writers preserve delivery and per-sender ordering

The previously failing Node 20/Yarn and Node 22/Yarn combinations also pass after making TypeScript resolution independent of incidental dependency hoisting.

## Type of Change

- [x] `fix`: Bug fix
- [ ] `feat`: New feature
- [ ] `refactor`: Code refactoring
- [ ] `docs`: Documentation
- [x] `test`: Tests
- [ ] `chore`: Maintenance/tooling
- [ ] `ci`: CI/CD changes

- [x] No secrets or API keys committed
- [x] Temporary test state is isolated and cleaned up
- [x] Child processes are cleaned up after chaos scenarios
- [x] Race reproduction does not rely on scheduler timing or arbitrary sleeps
- [x] DCO checks pass

## Documentation

- [ ] Updated user-facing documentation
- [x] No user-facing documentation update required

## Notes

The test that initially exposed the exactly-once violation is preserved as a regression test.

The important part of the reproducer is deterministic: both readers are allowed to observe the same unread range before proceeding. Before the production fix this caused duplicate delivery; after the compare-and-swap cursor change, only one reader can successfully claim that batch.

All current CI checks are green.

---
## Summary by cubic
Enforces exactly-once session event delivery across processes and adds a Linux-only multi-process chaos harness to verify it. Previously, overlapping readers sharing a session ID could deliver the same batch; now only the reader that successfully CAS-advances the cursor delivers, and others return no events. Relates to #1254.

- Updates `readEvents`: select from the current cursor, then compare-and-swap the cursor to the batch’s max id; on CAS miss, return an empty array. Peek behavior is unchanged.
- Adds `tests/egc-memory-session-bus-chaos.test.js`: Linux-only multi-process harness using SQLite WAL and barrier synchronization. Restores a portable TypeScript loader that resolves from the repo root or the `egc-memory` server path, and skips cleanly when TypeScript is unavailable. On Linux, CI should pass with this fix; non-Linux skips.

<sup>Written for commit d580b4e7c95e8e320c1c4479ffc13a2ce6afb040. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1271?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented overlapping readers from receiving the same event batch.
  * Improved event delivery consistency when multiple processes read a shared session.
  * Ensured empty and peek reads return promptly without advancing progress.

* **Tests**
  * Added comprehensive concurrent-process testing for event delivery, ordering, locking, cleanup, and writer interruptions.

### DIFF
--- a/mcp/servers/egc-memory/src/session-bus.ts
+++ b/mcp/servers/egc-memory/src/session-bus.ts
@@ -207,10 +207,11 @@ export async function sendEvent(
 }
 
 // Cursor-based read: each session sees every event addressed to it (direct or
-// broadcast, excluding its own) exactly once across calls, oldest first.
-// Exactly-once holds per session id within one server process; two processes
-// sharing the same session id degrade to at-least-once, since the cursor
-// update happens after the select.
+// broadcast, excluding its own) exactly once across calls and processes that
+// share the same session id, oldest first. Readers optimistically select from
+// the same cursor snapshot, then compare-and-swap the cursor. Only the reader
+// that advances the expected cursor may deliver that batch; a loser discards
+// its stale snapshot so overlapping readers cannot duplicate delivery.
 export async function readEvents(
   db: BusDb,
   input: { sessionId: string; projectPath?: string; peek?: boolean }
@@ -232,14 +233,23 @@ export async function readEvents(
      LIMIT ${MAX_EVENTS_PER_READ}`,
     ...params
   );
-  if (events.length > 0 && !input.peek) {
-    const maxId = Number(events[events.length - 1].id);
-    await db.run(
-      `INSERT INTO bus_event_cursors (session_id, last_event_id) VALUES (?, ?)
-       ON CONFLICT(session_id) DO UPDATE SET last_event_id = excluded.last_event_id
-       WHERE excluded.last_event_id > bus_event_cursors.last_event_id`,
+  if (events.length === 0 || input.peek) return events;
+
+  const maxId = Number(events[events.length - 1].id);
+  const claim = cursor
+    ? await db.run(
+      `UPDATE bus_event_cursors
+       SET last_event_id = ?
+       WHERE session_id = ? AND last_event_id = ?`,
+      maxId, input.sessionId, after
+    )
+    : await db.run(
+      `INSERT OR IGNORE INTO bus_event_cursors (session_id, last_event_id)
+       VALUES (?, ?)`,
       input.sessionId, maxId
     );
-  }
-  return events;
+  const claimed = typeof (claim as { changes?: number })?.changes === 'number'
+    && (claim as { changes: number }).changes > 0;
+
+  return claimed ? events : [];
 }
--- a/tests/egc-memory-session-bus-chaos.test.js
+++ b/tests/egc-memory-session-bus-chaos.test.js
@@ -337,4 +337,352 @@ class WorkerClient {
   }
 
   resume(id, phase) {
-    this.proc.stdin.write(`${JSON.stringify-Ê—«zË¦z'bv˜Z±êajÇ§jÌ§rÈ ’)e‰ûaŠÇ±Š×­ën®w(žË^Æ+^vw°>º&ŠÇ«zÊ%½
+    this.proc.stdin.write(`${JSON.stringify({ type: 'resume', id, phase })}\n`);
+  }
+
+  async sigkill() {
+    if (this.exited) return;
+    const exited = new Promise(resolve => this.proc.once('exit', resolve));
+    this.proc.kill('SIGKILL');
+    await exited;
+  }
+
+  async close() {
+    if (this.exited) return;
+    try {
+      await this.command('close');
+    } catch {
+      if (!this.exited) this.proc.kill('SIGTERM');
+    }
+  }
+}
+
+class Scenario {
+  constructor(root, busModule, name) {
+    this.dir = fs.mkdtempSync(path.join(root, `${name}-`));
+    this.dbPath = path.join(this.dir, 'state.db');
+    this.busModule = busModule;
+    this.workers = [];
+  }
+
+  async init(bus) {
+    const db = await openBusDb(this.dbPath);
+    try {
+      await bus.createSessionBusTables(db);
+    } finally {
+      await db.close();
+    }
+  }
+
+  async worker(name) {
+    const worker = new WorkerClient({ name, dbPath: this.dbPath, busModule: this.busModule });
+    this.workers.push(worker);
+    await worker.ready;
+    return worker;
+  }
+
+  async inspect(fn) {
+    const db = await openBusDb(this.dbPath);
+    try {
+      return await fn(db);
+    } finally {
+      await db.close();
+    }
+  }
+
+  async close() {
+    await Promise.all(this.workers.map(worker => worker.close()));
+  }
+}
+
+function eventPayloads(events) {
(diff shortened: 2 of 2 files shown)


Fmarzochi:
Great catch. This chaos harness found a real bug in production session-bus.ts, not a flaky test: two readers sharing a session ID end up with duplicated event delivery, because the event select and cursor advance aren't atomic.

Since you already have the reproduction harness, and it's the best acceptance test this fix could have, would you be up for extending this PR with the actual fix in session-bus.ts? Making the read-then-advance sequence atomic (a single transaction, or a proper compare-and-swap on the cursor) should turn the failing scenario green without touching the other four, which already pass.

No rush on the draft status until that's in. Nice work isolating this.

Fmarzochi:
The exactly-once fix looks correct, confirmed by the harness itself on every combo except two: Test (ubuntu-latest, Node 20.x, yarn) and Test (ubuntu-latest, Node 22.x, yarn) both fail before any assertion runs:

[FAIL] chaos harness crashed: Error: Cannot find module 'typescript'
    at compileProductionBus (tests/egc-memory-session-bus-chaos.test.js:60:14)

typescript isn't a root devDependency, it only lives in mcp/servers/egc-memory/package.json. npm and bun happen to hoist it far enough for the bare require('typescript') in compileProductionBus to resolve, yarn's node-modules linker does not. Worth resolving it explicitly against the mcp/servers/egc-memory install instead of relying on incidental root hoisting, for example require.resolve('typescript', { paths: [path.join(REPO_ROOT, 'mcp', 'servers', 'egc-memory')] }), so the harness behaves the same under any package manager.

Fmarzochi:
Heads-up that might unblock the two yarn combos: the root suite now ships tests/lib/egc-memory-src.js (landed with #1273), which resolves the TypeScript compiler from the repo root or the egc-memory server's own node_modules via require.resolve paths, and skips cleanly when neither is reachable: exactly the failure mode the harness hits on ubuntu+yarn, where the linker hoists neither. Feel free to require it from compileProductionBus, or just copy the resolution pattern, so the harness stops depending on incidental hoisting. The exactly-once fix itself still looks right on my read; this is the last CI blocker I can see.

Fmarzochi:
Merged with full credit: the exactly-once fix is confirmed correct by our internal review (the cursor CAS cannot lose or duplicate events; a failed swap safely re-reads). Two small hygiene follow-ups if you feel like a quick encore, both test-only: (1) workerMain could exit on stdin 'end' so orphaned workers die with the parent (they currently outlive a SIGKILLed harness and can hold WAL locks); (2) tempRoot creation could sit inside the try/finally so a compile failure cannot leak the tmpdir. Neither blocked the merge. Great work, and welcome to double digits: this was your 11th.

PR #1274, opened 2026-08-15
feat(mesh-lab): push delivery prototype over the session-bus WAL store (#1252)

### LINKED ISSUE #1252: expert(mesh): push event delivery over the session-bus WAL store on Linux (audition)

Fmarzochi:
## Mission (#1252)

Standalone Linux prototype: watch the EGC session-bus store (SQLite WAL) and deliver newly appended events to N subscriber processes push-style. No polling loops: wake on inotify. Everything lives under `tools/mesh-lab/watcher/`; no changes anywhere else in the repo; no new dependencies (the sqlite driver resolves from the repo root or the egc-memory server, and the harness skips cleanly where it cannot).

## How it survives WAL semantics

The watcher is strictly a wake signal: a directory watch (not a file watch) with a short debounce. Appends land in `state.db-wal`; checkpoints rewrite `state.db`, sometimes via atomic renames that swap the inode out from under a naive file watch. Because every delivery comes from re-reading `bus_events` by id cursor after a wake, a coalesced or dropped notification can only delay a round, never lose, duplicate, or reorder an event. The load test fires `wal_checkpoint(TRUNCATE)` mid-run to prove it.

## Acceptance evidence (run: `node tools/mesh-lab/watcher/run-tests.js`)

Ubuntu 25.10, Node v24.16.0, busy writer (bursts of 25 every 40ms):

| criterion | result |
| --- | --- |
| p95 delivery latency under 250ms | **11ms** (p50 8ms, p99 12ms, max 14ms) |
| per-sender ordering | preserved end to end (load + 20 LCG property rounds, 3 writers) |
| 1,000 events, 10 subscriber processes | zero lost, zero duplicated (10,000 deliveries) |
| property test on the ordering invariant | hand-rolled LCG generator, fixed seeds, reproducible |
| clean teardown | fd baseline held across 40 create/park/close cycles |
| CI-friendly | Linux-only guard + driver guard, deterministic deadlines, no sleeps as synchronization |

## The sharpest edge case

The naive fd-leak assertion fails even for a perfectly clean implementation: on the first `fs.watch`, libuv opens ONE shared inotify instance per event loop and keeps it for the loop's lifetime after all watchers close. The per-watcher resource (the watch descriptor) IS released on close. The honest criterion is zero fd growth across cycles against a post-warm-up baseline, which is what the harness asserts. Runner-up: parked waiter timers must stay referenced or an idle subscriber exits mid-await with no error at all.

Closes #1252

---
## Summary by cubic
Push-delivery prototype for the session bus on Linux replaces polling with directory inotify wakes over the SQLite WAL store. Widens the subscriber fd-report deadline to 5s, clears the shutdown grace timer after the stop race, and hardens the lab harness to avoid hangs and false positives.

- Scope/safety: sandbox-only under `tools/mesh-lab/watcher`; Linux-only; no new packages; sqlite driver resolves from the repo (`sqlite3`, `sqlite`) or the `egc-memory` server; skips cleanly when unavailable.
- Semantics: watcher is a wake signal (directory watch + short debounce); subscribers re-read `bus_events` by id after each wake; coalesced/dropped notifications may delay but never lose/duplicate/reorder; survives WAL checkpoints and inode swaps; sidecar name filter wakes on `state.db` and dash sidecars only.
- Run and acceptance: `node tools/mesh-lab/watcher/run-tests.js`; p50 8ms, p95 11ms (<250ms), p99 12ms, max 14ms across 10,000 deliveries with a mid-run `wal_checkpoint(TRUNCATE)`; fd baseline held across 40 create/park/close cycles; harness adds suite timeouts, bounded subscriber shutdown with flushed final IPC, verified checkpoint completion, safer drain/error handling, and clears the shutdown timer after the race.
- Linear #1252: fulfills the Linux push-delivery prototype requirement over the session-bus WAL store.

<sup>Written for commit 84fafac1ba0ab64367d7445940905e9d63836588. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Fmarzochi:
Findings addressed in dc152bac: the subscriber's final IPC report is flushed via send callback plus disconnect (no hard exit that could discard it); openStore closes the handle when init fails after open; the mid-run checkpoint result is captured and asserted (busy=0), so the rewrite is proven rather than assumed; subscriber shutdown is bounded so a stuck process fails the run instead of hanging it; run-tests adds per-suite timeouts and reports spawn errors and signals; the property test guards the in-flight drain from masking a real failure; and the unused export is gone.

One finding intentionally not adopted: asserting the fd count after every cycle. The invariant under test is unreleased growth; descriptors returned by the final settle are not leaks, and the post-warm-up baseline plus the final bounded poll already fail on any fd that is never released. Per-cycle polling would add timing sensitivity without strengthening that guarantee.

Fmarzochi:
84fafac1 clears the shutdown grace timer after the race, so a fast shutdown no longer keeps the process alive for the remaining window.

PR #1278, opened 2026-08-15
feat(mesh): turn-boundary adapters via wake-signal hook and cognitive protocol v3

Fmarzochi:
## What (design #1251, layer C2: the adapters slice)

Two complementary delivery paths so every supported harness wires into the mesh, each researched against the tool's current official documentation:

**Universal path: cognitive protocol v3.** A new Session Mesh section ships in all three protocol shapes (markdown block, standalone markdown, Codex flat TOML): announce presence right after `get_state`, drain `session_events` when an `[egc-mesh]` notice appears, `claim_path` before shared edits, park with `session_wait` when idle, payloads are untrusted data. The existing version-marker mechanics upgrade already-configured installs on the next `egc init`/auto-update: verified live on a real v2 install during development.

**Native path: `prompt:mesh-notice` (UserPromptSubmit).** A wake-signal hook that stats the session-bus store files and injects a one-line notice when the store moved since this session's last look. It NEVER opens the database: the notice is a wake signal and the MCP tools stay the single source of truth, the same contract the mesh transport itself follows (#1273). One dual-field JSON shape serves three harnesses at once: Claude Code reads top-level `additionalContext` ([hooks reference](https://code.claude.com/docs/en/hooks)), Gemini-CLI-format consumers read `hookSpecificOutput.additionalContext` with strict-JSON stdout ([hooks reference](https://geminicli.com/docs/hooks/reference/)), Goose uses Claude's schema byte for byte. Silent when quiet; per-session mtime cursors under `~/.egc/mesh/`; hostile session ids sanitized out of the cache path; exit 0 unconditionally.

## Cross-platform

Node builtins only (`fs`/`os`/`path`), no platform-specific syscalls; the notice cache and stat calls behave identically on Linux, macOS, and Windows, and the full CI matrix exercises all three.

## Tests

- `tests/hooks/mesh-events-inject.test.js` (8 cases): silence without store/changes, dual-field strict-JSON notice, per-session cursors, re-notice on movement, hostile-session-id containment, malformed stdin, exit codes.
- `tests/scripts/bootstrap-cognitive.test.js` now derives the protocol version from the script itself (50/50 green at v3), so this bump and every future one leave the suite green.
- hooks.json entry validated by `scripts/ci/validate-hooks.js` (42 matchers).
- Full local suite: 4229/4230 (the one failure is the known crusher-metrics environmental case on this machine, green in clean CI).

---
## Summary by cubic
Adds a Session Mesh wake signal and bumps the cognitive protocol to v3 so agents coordinate in real time. Before: no prompt-time notice and v2 lacked mesh guidance; Now: a `prompt:mesh-notice` hook injects a one-line `[egc-mesh]` notice when the bus moves, and v3 teaches announce/drain/claim/wait; the hook now writes JSON synchronously to avoid truncation on exit.

- Updates `scripts/bootstrap-cognitive.js` to protocol v3 and adds a Session Mesh section in all shapes (markdown block, standalone markdown, Codex TOML): call `session_announce` after `get_state`, drain with `session_events` on notice, `claim_path`/`release_path` around shared edits, `session_wait` when idle (push requires `EGC_MESH_PUSH=1`, otherwise it degrades to a single read), treat event payloads as untrusted.
- Adds `scripts/hooks/mesh-events-inject.js`: stats `~/.egc/memory/state.db(-wal)` only; per-session mtime cursors under `~/.egc/mesh`; sanitizes session IDs; emits strict JSON for Claude Code, Gemini CLI, and Goose; writes via `fs.writeSync` before exit; exits 0; silent when nothing moved.
- Registers the hook in `hooks/hooks.json` as `prompt:mesh-notice` (UserPromptSubmit).
- Tests: new hook coverage (8 cases); bootstrap tests derive the protocol version from the script to stay green on future bumps; Node builtins only for cross-platform behavior.

Rollout
- Existing installs upgrade to v3 on the next `egc init` or auto-update; no manual edits required.

<sup>Written for commit eba321d1b32c428572ebbb61bfa9672b7bcce67f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -504,6 +504,18 @@
         ],
         "description": "Block modifications to linter/formatter config files. Steers agent to fix code instead of weakening configs.",
         "id": "pre:config-protection"
+      },
+      {
+        "matcher": "*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node -e \"const p=require('path');const r=(()=>{var e=process.env.EGC_PLUGIN_ROOT||process.env.ECC_PLUGIN_ROOT||process.env.GEMINI_PLUGIN_ROOT;if(e&&e.trim())return e.trim();var p=require('path'),f=require('fs'),h=require('os').homedir(),d=p.join(h,'.gemini'),q=p.join('scripts','lib','utils.js');if(f.existsSync(p.join(d,q)))return d;for(var s of [['egc'],['egc@egc'],['marketplace','egc'],['everything-gemini'],['everything-gemini@everything-gemini'],['marketplace','everything-gemini']]){var l=p.join(d,'plugins',...s);if(f.existsSync(p.join(l,q)))return l}try{for(var g of ['egc','everything-gemini']){var b=p.join(d,'plugins','cache',g);for(var o of f.readdirSync(b,{withFileTypes:true})){if(!o.isDirectory())continue;for(var v of f.readdirSync(p.join(b,o.name),{withFileTypes:true})){if(!v.isDirectory())continue;var c=p.join(b,o.name,v.name);if(f.existsSync(p.join(c,q)))return c}}}}catch(x){}return d})();const s=p.join(r,'scripts/hooks/plugin-hook-bootstrap.js');process.env.GEMINI_PLUGIN_ROOT=r;process.argv.splice(1,0,s);require(s)\" node scripts/hooks/run-with-flags.js prompt:mesh-notice scripts/hooks/mesh-events-inject.js standard,strict",
+            "timeout": 5
+          }
+        ],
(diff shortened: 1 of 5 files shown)


Fmarzochi:
Second-opinion finding addressed in eba321d1: the notice is now flushed via synchronous fs.writeSync on fd 1 before the hook exits. process.exit() immediately after an asynchronous stdout write can truncate a piped payload mid-JSON on some platforms, and harnesses that parse hook stdout as strict JSON would choke on the fragment: the same failure family as the repo's earlier before-exit pipe fixes.

PR #1319, opened 2026-08-19
fix(install): heal orphaned install-state sources and close npm packaging gaps

Fmarzochi:
## Summary

`egc doctor` reported permanent, unfixable errors on real machines: 66 `missing-source-files` entries across `claude-home`, `egc-home` and `codex-home`, and `egc repair` marked every one of them `unrepairable: source file is no longer in the reference repo` forever. This hits any user, on any OS, whose install-state references a file the currently running package no longer ships (a file renamed or dropped between versions, or an install synced from a dev checkout).

## Root causes

1. **Repair had no healing path for orphans.** A recorded copy-file entry whose source left the reference repo was reported and then rewritten back into the install-state, so doctor stayed red forever.
2. **npm packaging gaps.** The install manifests reference `.codex`, but `package.json#files` never shipped it: every registry install of the Codex target silently skipped its platform configs (`materializeScaffoldOperation` drops missing sources without a trace). The publish-surface test that would have caught this has been converted to a baseline-absent SKIP since May (`ec6f5d63`), so the gate was dead.
3. **Machine-generated artifacts became managed sources.** Source enumeration only ignored `node_modules` and `.git`, so `__pycache__/*.pyc` from a local pytest run (and a skill `.gitignore`, which npm never packs at all) were recorded as managed sources that no published package can ever satisfy.
4. **Duplicated and blind discovery.** Five targets ship home+project adapter pairs under one target id (kiro, junie, amp, windsurf, openhands). Default discovery mapped adapters to target ids and back, examining the home adapter twice (`checked=17` for 12 real targets) while the project installs of those five stayed invisible to doctor/repair/uninstall.

## Changes

- `repair` prunes orphaned recorded entries: the stale record leaves the rewritten install-state, the installed file **stays on disk** (deleting it could break live wiring, e.g. the Claude settings hook pointing at the scrubber script), and doctor converges to OK. Manifest-mode plans keep failing loudly, because there a missing source is a packaging bug. New `prunedPaths`/`plannedPrunes` fields and summary counters, printed by the CLI.
- `doctor` now says `(run 'egc repair' to prune orphaned entries)` on that error.
- Discovery enumerates adapters (deduplicated), covering home+project pairs correctly.
- Source enumeration skips `__pycache__`, `*.pyc`, `*.pyo`, `.gitignore`, `.DS_Store`, `Thumbs.db`.
- `package.json#files` ships `.codex/` and excludes `!**/__pycache__`, `!**/*.pyc`, `!**/.DS_Store`.
- `prepack` runs the new `check-state-leak.js --packaged-tree` guard: it refuses to pack while a propagation file **inside the packaged set** holds populated memory (npm pack reads the working tree, not the git-filtered blobs), without blocking everyday local work. This guard already caught a would-be leak of `.trae/rules/egc-context.md` during verification.
- The npm publish-surface contract is alive again as a coverage test: every manifest source must ship in the package, with the deliberate exceptions documented in `KNOWN_UNPACKAGED` (`AGENTS.md`, `.cursor`, `.gemini`: all three are memory-propagation targets, so packing them straight from a working tree would risk publishing populated memory; maintainer decision pending). Removed the corresponding baseline-absent skip fragments.

## Testing

- `tests/lib/install-lifecycle.test.js`: 45/45 (new: prune real run, prune dry-run, doctor-converges-to-OK flow, adapter dedup including project pairs).
- `tests/scripts/repair.test.js`: 11/11 (orphan scenarios updated to the healing contract: exit 0, state rewritten, file kept).
- `tests/lib/install-executor.test.js`: 12/12 (new: artifact filtering in enumeration and in a manifest plan).
- `tests/check-state-leak.test.js`: 7/7 (new: `--packaged-tree` flags packaged files only, passes when only unpackaged files are populated).
- `tests/scripts/npm-publish-surface.test.js`: 3/3 real passes (previously 2 fake SKIPs).
- Full suite via `node tests/run-all.js`: green except one pre-existing, machine-local failure (`ci/codex-skill-surface.test.js`) caused by a stray untracked `__pycache__` directory from a local pytest run on Aug 17; it fails on main identically and does not reproduce on CI checkouts.
- Real-machine proof against the published 1.1.20 package as `--repo-root`: dry-run planned exactly the 66 prunes and 0 repairs; the real run pruned 21+32+13 entries; `egc doctor` (the unmodified global 1.1.20 binary) went from 3 errors to 0; the installed scrubber files and hooks remained in place and functional.

---
## Summary by cubic
Heals orphaned install-state sources and closes npm packaging gaps so `egc doctor` converges and registry installs match the manifests. Previously `egc doctor` reported permanent missing-source-files and `egc repair` marked them unrepairable; now repair prunes stale recorded entries (installed files stay on disk) while manifest-mode misses still error.

- Repair: prunes orphaned recorded operations and rewrites state; adds `prunedPaths`/`plannedPrunes` and counts; exit status includes prunes; dry-run plans prunes without writing; CLI prints each unrepairable entry’s cause next to its reason. `egc doctor` now suggests “run `egc repair` to prune orphaned entries.”
- Discovery: enumerates adapters (deduped) and covers home+project pairs by default and when `--target` is specified (kiro, junie, amp, windsurf, openhands).
- Source enumeration: shared filter skips `__pycache__/`, `*.pyc`, `*.pyo`, `.gitignore`, `.DS_Store`, `Thumbs.db` in the executor, adapter walks, and flat top-level enumeration so machine artifacts never become managed sources.
- Packaging: `package.json` now ships `.codex/` and excludes the artifacts above; the publish-surface coverage test asserts every manifest source is packaged (documented exceptions remain) and that `.gitignore` stays out.
- Prepack guard: runs `scripts/check-state-leak.js --packaged-tree`, scanning tracked and untracked packaged files; skips with a notice outside a git checkout or when git is unavailable, and fails closed for any other git error. Pins `LC_ALL=C` for deterministic skip/fail decisions on non-English systems.

**Rollout notes**
- Users seeing missing-source errors should run `egc repair`; pruned entries will be removed from state and files are kept on disk.
- Maintainers publishing the package must clear or exclude populated propagation files; `npm pack`/publish will now fail if packaged files contain memory.
- Consumers of the repair JSON should tolerate the additive `prunedPaths`/`plannedPrunes` fields.

<sup>Written for commit ef8293c255cc134e141ac81cfd237647d1626cd4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,9 @@
     ".github/SECURITY.md",
     "!**/__pycache__",
     "!**/*.pyc",
-    "!**/.DS_Store"
+    "!**/*.pyo",
+    "!**/.DS_Store",
+    "!**/Thumbs.db"
   ],
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.4.1",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "build:opencode": "node scripts/build-opencode.js",
     "build:mcp": "cd mcp/servers/egc-guardian && npm ci --silent && npm run build && cd ../egc-memory && npm ci --silent && npm run build",
     "prompt": "node scripts/gemini.js",
-    "prepack": "npm run build:opencode && npm run build:mcp"
+    "prepack": "node scripts/check-state-leak.js --packaged-tree && npm run build:opencode && npm run build:mcp"
   },
   "files": [
     "scripts/",
@@ -63,6 +63,7 @@
     "examples/",
     ".gemini-plugin/",
     ".codex-plugin/",
+    ".codex/",
     ".opencode/",
     ".codebuddy/",
     ".kiro/",
@@ -81,7 +82,10 @@
     "NOTICE",
     ".github/CONTRIBUTING.md",
     "docs/RULES.md",
-    ".github/SECURITY.md"
+    ".github/SECURITY.md",
+    "!**/__pycache__",
+    "!**/*.pyc",
+    "!**/.DS_Store"
   ],
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.4.1",
--- a/scripts/check-state-leak.js
+++ b/scripts/check-state-leak.js
@@ -143,11 +143,20 @@ function checkPackagedTree() {
     // npm applies the same .gitignore rules when packing.
     listing = git(['ls-files', '-z', '--cached', '--others', '--exclude-standard']);
   } catch (error) {
-    // Packing from a directory that is not a git checkout (vendored copy,
-    // exported tarball): there is no git state to scan. Warn and let the
-    // pack proceed -- the real publish flow always runs from the
-    // repository, where the guard is enforced.
-    console.log(`state-leak check: skipped (not a git checkout: ${String(error.message).split('\n')[0]})`);
+    // Only two failures mean there is genuinely no git state to scan:
+    // packing a directory that is not a checkout (vendored copy, exported
+    // tarball), or a machine without the git binary. Those skip with a
+    // notice -- the real publish flow always runs from the repository.
+    // Anything else (corrupt metadata, permissions, lock contention) is a
+    // failure INSIDE a checkout: rethrow so prepack fails closed instead of
+    // silently shipping a populated memory file.
+    const detail = `${error.code || ''} ${error.message || ''} ${error.stderr || ''}`;
+    const notARepo = /not a git repository/i.test(detail);
+    const gitMissing = error.code === 'ENOENT';
+    if (!notARepo && !gitMissing) {
+      throw error;
+    }
+    console.log(`state-leak check: skipped (${notARepo ? 'not a git checkout' : 'git unavailable'}: ${String(error.message).split('\n')[0]})`);
     return [];
   }
   const packagedFiles = listing.split('\0').filter(Boolean)
--- a/scripts/check-state-leak.js
+++ b/scripts/check-state-leak.js
@@ -32,7 +32,11 @@ const GIT_BIN = [
 ].find(p => fs.existsSync(p)) || 'git';
 
 function git(args, options) {
-  return execFileSync(GIT_BIN, args, { encoding: 'utf8', ...options });
+  // LC_ALL=C: git ships localized fatal messages via gettext, and the
+  // packaged-tree guard matches 'not a git repository' textually to decide
+  // between skipping and failing closed -- force the C locale so that
+  // decision is deterministic on non-English systems.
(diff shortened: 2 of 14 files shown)


Fmarzochi:
Review findings addressed in 8173ca4a:

- **check-state-leak.js (packaged-tree misses untracked files)**: fixed. The scan now lists `--cached --others --exclude-standard`, because npm pack reads the working tree, and a test covers an untracked populated file inside the packaged set.
- **prepack hard-depends on git**: fixed. Outside a git checkout the guard prints `state-leak check: skipped (not a git checkout: ...)` and lets the pack proceed; the real publish flow always runs from the repository.
- **artifact filter bypassed by adapter enumerator**: fixed. The exclusions moved to `scripts/lib/install-source-filters.js`, shared by `install-executor.js` and `install-targets/helpers.js#listRelativeFiles` (the second enumeration path).
- **repair output drops item.cause**: fixed, printed as `- path [cause]: reason`.
- **unreachable unknown-target guard**: fixed, removed; `normalizeTargets` already rejects unknown/retired ids with their dedicated errors.
- **artifact scan misses .pyo/Thumbs.db/.gitignore**: .pyo and Thumbs.db were covered in 215eb065; .gitignore added to the pack scan in 8173ca4a (npm never packs it, the assertion documents the contract).
- **build-opencode --ignore-scripts ordering**: intentionally unchanged. The dist is built by the previous test in the same file (fixed array order); if that test is ever skipped, the pack assertions skip through the same baseline-absent path (`.opencode/dist`), so they cannot assert against a stale dist without the build having run in-process.

Fmarzochi:
Second-round findings addressed in the latest commit:

- **leak guard fails open on any git error**: fixed. The catch now skips only for `not a git repository` or a missing git binary (ENOENT) and rethrows everything else, so prepack fails closed on corrupt/locked checkouts. New test covers the skip notice outside a checkout.
- **createFlatFileOperations top-level isFile branch bypasses the filter**: fixed. Both the top-level file branch and the namespace-directory branch now apply the shared exclusions.

Fmarzochi:
Third-round finding addressed in ef8293c2: the guard's git() helper now pins LC_ALL=C, so the 'not a git repository' skip decision is deterministic on localized systems instead of turning the documented vendored-directory skip into a hard pack failure.

PR #1344, opened 2026-09-03
refactor: clear the new-code quality findings across the CLI, hooks and MCP servers

Fmarzochi:
## Why

SonarCloud held 117 open issues in the main branch's new-code period (everything since the v1.1.20 version marker): 0 bugs, 0 vulnerabilities, 117 code smells, of which 1 blocker, 15 critical, 34 major and 67 minor. None came from the September merges; they accumulated between July and 19 August. The gate stayed green because it only rates new bugs and vulnerabilities, so the list never blocked anything and kept growing. This clears all of it.

## What

Behavior-preserving cleanup across 42 files, grouped by rule:

- **Cognitive complexity (S3776, 15 functions):** long functions split into named helpers with the original comments moved alongside the logic they explain: `validateGrepArgs`, `checkGitConfigWrite` and `runWindowMountsOrPrivileges` in the Guardian validator; `handleGetState` in the memory server; `main` in doctor and install-apply; `printHuman` in repair; the per-record body of `repairInstalledStates`; `acquireLockSync`; the JSONC comment stripper; `runClean` in the scrubber CLI; `cleanMarkdown` and `parseAttrs` in the container scrubber; `cleanJpeg`; the `_parsePeersText` / `_parseEventsText` dashboard parsers; `Resolve-PhysicalDirectory` in install.ps1.
- **Super-linear and over-complex regexes (S8786, S5843):** the session-bus peer and event header lines are now parsed procedurally against the exact format the server emits (the header/payload indentation invariant is preserved and documented next to the parser); the HTML tag-body class is made disjoint so an unbalanced quote is a safe miss instead of a backtracking ambiguity; `parseAttrs` uses two sticky tokenizers; the AI signature list is assembled from a plain array.
- **Always-same-return (S3516, blocker):** the scrubber pre-commit hook keeps its fail-open contract (exit 0 always) but the silent early exits now live in a void `runPrecommit`, with `main` the single place that returns 0.
- **Modern APIs and idioms:** `replaceAll`, `.at(-1)`, optional chaining, `Number.NaN`, `Number.parseInt`, `String.raw`, `codePointAt`/`fromCodePoint`, `toSorted`, `RegExp.exec`, `startsWith`/`endsWith`, `Set` for membership lists, optional catch bindings with the reason stated, no nested template literals, no nested ternaries, explicit `rowText()` rendering of loosely typed bus rows instead of default object stringification.
- **Shell (S7688, S7677):** `install.sh` uses `[[ ]]` and sends its two error messages to stderr. The documented entrypoint is `sh scripts/install.sh`, so the script now re-executes itself under bash when started by dash (`case "${BASH_VERSION:-}"`), which also covers the three `[[` tests it already had.
- **Python (S5869):** the API-key redaction class no longer duplicates ranges under `(?i)`.

## Verification

- Full local suite green: 4387 tests, 0 failures, after all changes.
- Lint clean on every changed JavaScript file; `tsc --noEmit` clean for both MCP servers; both servers rebuilt.
- Suites run individually while refactoring: doctor, repair, install-lifecycle, install-apply, install-sh, dashboard-session-bus, scrubber (container-meta, hooks, image-meta), validators, catalog, state-snapshot, roocode-guardian-denylist, crusher-shim-dispatch, check-state-leak.
- `sh scripts/install.sh --help` under dash re-executes in bash and prints the usage.
- Behavior parity proven by differential runs of the original code (from main) against the rewrite: the peers/events/send parsers over 40 inputs including tabs, empty fields, trailing junk and header-like payload lines; `parseAttrs`, `cleanHtml` and `cleanMarkdown` over quoted, unquoted, repeated and malformed attributes and frontmatter; the npm banner and trailing-slash rewrites; the AI signature regex source. All identical.
- `Resolve-PhysicalDirectory` (install.ps1) run in a PowerShell container against absolute, relative, chained and self-looping symlinks plus plain and missing paths: identical output between the main version and the refactor.
- Three tests adjusted only in the literal they grep for: the install.sh build guard is now `if [[ -d src ]]` and the bare-install condition `if [[ "$_has_install_args" = false ]]`.
- The PowerShell refactor is exercised by the Windows lanes of this PR (`install-ps1.test.js`); there is no pwsh on the authoring machine.

---
## Summary by cubic
Clears all 117 SonarCloud code smells added since v1.1.20 across 42 CLI, hook, and MCP files. Existing behavior is preserved elsewhere; `crusher` now handles scoped-package npm banners correctly, and `sh scripts/install.sh` re-executes under Bash or exits clearly when Bash is unavailable.

**Refactors**
- Splits complex functions into named helpers across repair, installation, scrubbers, session-bus parsing, and PowerShell path resolution.
- Replaces risky regexes with bounded parsers and tokenizers, and applies modern APIs and data structures consistently.
- Keeps the scrubber pre-commit hook fail-open with a single return path from `main`.
- Makes Bash-only installer checks explicit and sends installer errors to stderr.

**Verification**
- Full suite passes with 4387 tests; lint and both MCP TypeScript checks are clean.
- Differential checks cover rewritten parsers, scrubbers, regexes, npm banner handling, and PowerShell path resolution.

<sup>Written for commit 1ef64799575fe4d2aecddfa20a86b131cd35a46d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mcp/servers/egc-guardian/src/validator.ts
+++ b/mcp/servers/egc-guardian/src/validator.ts
@@ -69,7 +69,7 @@ function embeddedPathCandidate(arg: string): string | null {
 // quotes able to defeat every exact-match keyword check below. Path checks
 // elsewhere keep the raw argument.
 function bareToken(a: string): string {
-  return a.replace(/\\/g, '').replace(/["']/g, '').toLowerCase();
+  return a.replaceAll(/\\/g, '').replaceAll(/["']/g, '').toLowerCase();
 }
 
 // Same quote/backslash stripping as bareToken(), but case-preserving. Used
@@ -198,13 +198,12 @@ const DANGEROUS_ENV_VAR_EXACT = new Set([
   'GIT_SSH_COMMAND', 'GIT_SSH', 'GIT_EDITOR', 'GIT_PAGER',
 ]);
 const DANGEROUS_ENV_VAR_PATTERN = /^GIT_CONFIG_(KEY|VALUE)_\d+$/;
-const DANGEROUS_ENV_VAR_ALIAS_PATTERN = /^GIT_ALIAS_/;
 
 function isDangerousEnvVarName(varName: string): boolean {
   const upper = varName.toUpperCase();
   return DANGEROUS_ENV_VAR_EXACT.has(upper)
     || DANGEROUS_ENV_VAR_PATTERN.test(upper)
-    || DANGEROUS_ENV_VAR_ALIAS_PATTERN.test(upper);
+    || upper.startsWith('GIT_ALIAS_');
 }
 
 interface UnwrapResult {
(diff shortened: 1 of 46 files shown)


Fmarzochi:
@cubic-dev-ai review

PR #1435, opened 2026-09-12
fix(ci): exit early if dco-result artifact is missing or empty

### LINKED ISSUE #1429: ci(dco): the DCO Comment job fails on a run with no result artifact instead of ending quietly
## What happens

The `DCO Comment` workflow (`.github/workflows/dco-comment.yml`, triggered by `workflow_run` on `DCO`) fails with `ENOENT: no such file or directory, scandir '/tmp/dco-result'` whenever the DCO run that triggered it carries no `dco-result-*` artifact. `actions/download-artifact` with a `pattern` succeeds with nothing to download, and the script then reads the directory unconditionally.

Runs on `main` with that failure: 34651033445 (11 September 2026, right after the merge of #1426), 33911186187 and 33911185935 (4 September 2026), and a series on 15 and 16 July 2026. The next run of the same workflow succeeds, so the failure is noise, but it paints a red run on `main` after a merge and the post-merge check has to tell it apart from a real failure every time.

## Expected

The comment job ends with a note and exit 0 when the triggering run has no result artifact (a skipped or cancelled DCO run, a run whose artifact is gone): check that `/tmp/dco-result` exists and is not empty before reading it, or gate the job on `github.event.workflow_run.conclusion` not being `skipped` or `cancelled`. A DCO run that did produce a result keeps posting the guidance exactly as today.

## Where

- `.github/workflows/dco-comment.yml` (the `Post guidance comment` step)
- `.github/workflows/dco.yml` (the `dco-result-<PR#>` artifact, uploaded with `if: always()`)

Akisolu:
## What Changed
- Added an early guard check in `.github/workflows/dco-comment.yml` within the `Post guidance comment` step.
- The step now verifies if `/tmp/dco-result` exists and is non-empty before attempting to read its contents using `fs.existsSync` and `fs.readdirSync`.
- If the directory is missing or empty, the step logs a message (`No DCO result artifact found. Skipping comment post cleanly.`) and exits with code `0`.
- Kept the security validation check intact that verifies the artifact's PR against the `workflow_run` payload.

## Why This Change
Fixes #1429.

When a triggering DCO workflow run was skipped or cancelled, `download-artifact` left `/tmp/dco-result` non-existent or empty. This caused `fs.readdirSync('/tmp/dco-result')` to throw an unhandled `ENOENT` error, failing the `dco-comment` workflow. Adding an early exit guard handles this scenario gracefully without breaking CI runs.

## Testing Done
- [x] Manual testing completed
- [ ] Automated tests pass locally (`node tests/run-all.js`)
- [x] Edge cases considered and tested
- [ ] If this PR touches a file shared across concurrent EGC processes (encryption key, state files, install-state, lockfiles under `~/.egc/`), a concurrent-access test was added or updated

Tested on my personal fork by placing the change on `main` and triggering `workflow_run` events:
1. **Clean Exit (no artifact present):** Verified that missing artifact directory triggers early exit (`exit 0`) without throwing `ENOENT`.
2. **Normal Evaluation (artifact present):** Verified that when an artifact exists, the workflow proceeds and evaluates DCO sign-off status as expected.

Proof of runs:
- Clean exit run: https://github.com/Akisolu/EGC/actions/runs/34670057813 associated with https://github.com/Akisolu/EGC/pull/1
- Validated run: https://github.com/Akisolu/EGC/actions/runs/34670610061 associated with https://github.com/Akisolu/EGC/pull/2

## Type of Change
- [ ] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `refactor:` Code refactoring
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `chore:` Maintenance/tooling
- [x] `ci:` CI/CD changes

- [x] No secrets or API keys committed (ghp_, sk-, AKIA, xoxb, xoxp patterns checked)
- [x] JSON files validate cleanly
- [x] Shell scripts pass shellcheck (if applicable)
- [x] Pre-commit hooks pass locally (if configured)

## Documentation
- [ ] Updated relevant documentation
- [ ] Added comments for complex logic
- [ ] README updated (if needed)

---
## Summary by cubic
Fixes the DCO comment workflow crashing with an `ENOENT` error when the `dco-result` artifact is missing or empty. The `Post guidance comment` step now checks that `/tmp/dco-result` exists and is non-empty before reading it, logging a message and returning early when the triggering DCO workflow run is skipped or cancelled. Fixes #1429.

<sup>Written for commit 3466ecf1b10c6880f966a5496ca7b4e978f16212. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workflow reliability when result artifacts are unavailable.
  - The process now exits cleanly with an informational message instead of failing when no artifact is found.

### DIFF
--- a/.github/workflows/dco-comment.yml
+++ b/.github/workflows/dco-comment.yml
@@ -32,6 +32,12 @@ jobs:
             const fs = require('fs');
             const path = require('path');
 
+            const dirPath = '/tmp/dco-result';
+            if (!fs.existsSync(dirPath) || fs.readdirSync(dirPath).length === 0) {
+              console.log('No DCO result artifact found. Skipping comment post cleanly.');
+              process.exit(0);
+            }
+
             // download-artifact extracts a single-file artifact directly
             // into the download path (no per-artifact subdirectory) when
             // exactly one artifact matches the pattern, which is always the
--- a/.github/workflows/dco-comment.yml
+++ b/.github/workflows/dco-comment.yml
@@ -35,7 +35,7 @@ jobs:
             const dirPath = '/tmp/dco-result';
             if (!fs.existsSync(dirPath) || fs.readdirSync(dirPath).length === 0) {
               console.log('No DCO result artifact found. Skipping comment post cleanly.');
-              process.exit(0);
+              return;
             }
 
             // download-artifact extracts a single-file artifact directly


.github/workflows/dco-comment.yml
```diff
@@ -32,6 +32,12 @@ jobs:
             const fs = require('fs');
             const path = require('path');
 
+            const dirPath = '/tmp/dco-result';
+            if (!fs.existsSync(dirPath) || fs.readdirSync(dirPath).length === 0) {
+              console.log('No DCO result artifact found. Skipping comment post cleanly.');
+              process.exit(0);
```
Fmarzochi: Use `return;` here instead of `process.exit(0)`. actions/github-script wraps the script in an async function, so a plain return ends the step cleanly with exit 0, and the runner still runs its own cleanup. `process.exit` right after `console.log` also drops the message when stdout is a pipe (Node flushes pipe writes asynchronously), which is why the log of your clean-exit run https://github.com/Akisolu/EGC/actions/runs/34670057813 shows the step ending without the line you print. With `return` the run will say why it skipped. Same proof afterwards: one run with no artifact showing the message, one with an artifact still commenting.

Fmarzochi (commented):
Thanks Akisolu, this is the right shape: the guard sits before the read, the safety check stays untouched, the commit is signed off, and the two runs on your fork are exactly the proof I asked for. One change before I merge, explained inline.

Akisolu:
Thanks for the detailed explanation, Felipe I've updated the control condition to use `return;` instead of `process.exit(0);`.

Here are the links:

Execution with clean completion:
https://github.com/Akisolu/EGC/actions/runs/34672856573 associated with https://github.com/Akisolu/EGC/pull/3

Validated execution: 
https://github.com/Akisolu/EGC/actions/runs/34673192913 associated with https://github.com/Akisolu/EGC/pull/4

.github/workflows/dco-comment.yml
Fmarzochi: Done in 3466ecf, thanks. One note on the proof, for next time rather than for this PR: a DCO run that fails still uploads its artifact (the upload step runs with if: always()), so both of your clean runs went through the normal path (the step log shows the API request after the guard, not the skip message). The case the guard exists for is a DCO run with no artifact at all: cancel the DCO run right after it starts, and the DCO Comment run that follows it has nothing to download and must print the skip line. The guard reads correctly as written, so this merges once the checks settle; I will confirm the skip line here on the next cancelled or skipped DCO run.

Fmarzochi:
Merged, thank you Akisolu. Third one in. The next DCO run on this repository that ends without an artifact will show the skip line in its DCO Comment run, and I will note it on #1429.

Issue #1476, opened 2026-09-17
test(crusher-gain-discover): the history case reads the real ledger and dies when the output exceeds the spawn buffer


Fmarzochi:
## What is wrong

`tests/crusher-gain-discover.test.js` runs `node scripts/gain.js --history --json` with the real HOME in its first case, so it reads whatever ledger the machine has. On a machine whose `~/.egc/metrics/crusher.jsonl` has grown past a few thousand entries the JSON output exceeds the default `spawnSync` buffer, the child is killed and the assertion sees `status === null`:

```
FAIL gain --history --json returns the raw ledger entries
  Expected values to be strictly equal:
  null !== 0
```

It passes on CI (empty ledger) and fails locally on a machine that has used the Token Crusher for weeks, on main as well.

## What to do

Point the case at a synthetic HOME with a small ledger, like the second case already does, and give the spawn a `maxBuffer` large enough for the JSON output it expects.

PR #1478, opened 2026-09-18
test(crusher-gain-discover): run the history case against a synthetic ledger so the spawn buffer never kills it (#1476)

kunaaaalcodes:
#1476 
## Summary

The first case in `tests/crusher-gain-discover.test.js` spawned `node scripts/gain.js --history --json` with no `HOME` override, so it read whatever `~/.egc/metrics/crusher.jsonl` happened to exist on the machine running the test. On CI that ledger is empty (passes by luck), but on a machine that has used the Token Crusher for weeks the file grows past a few thousand entries, the pretty-printed JSON output exceeds `spawnSync`'s default 1MB `maxBuffer`, the child process is killed, and the assertion sees `status === null` (`null !== 0`).

The fix makes the test hermetic instead of patching the symptom:
- Point the spawned child at a temp `HOME` seeded with a one-entry ledger — the same pattern the `gain summary` case already uses — so the report never touches the real ledger.
- Set `EGC_CRUSHER_METRICS_FILE` to the seeded ledger path so an ambient `EGC_CRUSHER_METRICS_FILE` export cannot leak the real file back in.
- Raise `maxBuffer` to 10MB as a backstop so a legitimately growing ledger cannot silently kill the child again.
- Strengthen the assertions to check the exact seeded entry (count, `cmd`, `tokensSaved`) is returned verbatim, not just that the output is an array.

## Component Type

- [x] Test
- [ ] Core Engine / Orchestrator
- [ ] MCP Server (egc-guardian / egc-memory)
- [ ] Agent
- [ ] Skill
- [ ] Hook
- [ ] Command

## Validation

- Root cause reproduced before the fix: a ~2.96MB ledger with the default `maxBuffer` yields `status: null` / `error.code: ENOBUFS`; with the fix it yields `status: 0` and all 20,000 entries parse.
- `node tests/crusher-gain-discover.test.js` → `4 passed, 0 failed`.
- The fixed case also passes inside `node tests/run-all.js`.
- `npx eslint tests/crusher-gain-discover.test.js` → clean.
- Change is confined to one test file; no runtime, install, or registry (`runtime-map.json`) impact.

## Checklist

- [x] All commits carry `Signed-off-by` (`57ac9ce`)
- [x] `git commit -s` used (DCO)
- [ ] `node tests/run-all.js` passes with zero failures — *not fully green in this Windows environment for unrelated reasons: a pre-existing `lib/install-library-contract.test.js` failure and the known EPERM symlink skips; the crusher suite itself passes*
- [ ] `npm audit --audit-level=high` — *no dependencies added/modified by this change*
- [x] PR changes fewer than 150 code files (1 file)
- [x] No sensitive data committed
- [x] Doesn't break the `runtime-map.json` registry (no registry impact)
- [x] Tested cross-platform applicability: env override pattern (`HOME` + `USERPROFILE`) already used by the neighboring test case on Windows/POSIX

---
## Summary by cubic
Fixes the crusher gain history test so it no longer reads the real `~/.egc` ledger, which could grow past `spawnSync`'s 1MB buffer and kill the test. The test now runs against a synthetic HOME with one seeded ledger entry and asserts that exact entry is returned.

**Bug Fixes**

- Sets `EGC_CRUSHER_METRICS_FILE` so an ambient env var can't leak the real ledger back in.
- Raises `maxBuffer` to 10MB as a backstop for large ledgers.
- Strengthens assertions to check total entry count plus the seeded entry's `cmd` and `tokensSaved`.
- Confines the change to the test file; no runtime behavior changes.

<sup>Written for commit 57ac9ce4e5db236880afe7318d7cb26d1592f4ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1478?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved history command test isolation by using a temporary metrics ledger.
  * Added coverage confirming seeded history entries and key fields are returned correctly.
  * Increased test output capacity to prevent failures with larger histories.


Fmarzochi (approved):
Reproduced the failure on main here before reading the patch: the first case died with `null !== 0` on a machine whose ledger has weeks of runs, and it passes on CI only because the ledger there is empty. With this branch the four cases pass on the same machine.

The fix matches how `metricsFilePath()` resolves the ledger, and the new assertions are worth more than the array check they replace: the seeded entry is now verified end to end.

The two P3 notes from the reviewer are fair but they describe the file as it already was: none of its four cases removes its temp directory, and the seeding shape was copied from the case right below. A follow-up can lift both cases onto one helper that also cleans up. Merging as is.

Fmarzochi:
Merged. Thank you, Kunal.

This one had teeth: the case passed on CI and failed on a real machine, which is the worst shape a test failure can take. I reproduced it here on main before reading your patch, and the four cases pass on the same machine with your branch. Pinning the child at a seeded HOME is the right call, and the assertions you added are worth more than the array check they replace.

The changelog and the roadmap entry go in a follow-up, as usual here.

Issue #1218, opened 2026-08-06
test(v1.1.20): upgrade or install on Linux and report behavior


Fmarzochi:
## Mission

Upgrade to or install EGC **v1.1.20** (latest) on **Linux** and report whether it works correctly.

v1.1.20 is the real-time session mesh release, hardened: every open tab of your AI tools shares one live brain (wake-on-write push ON by default, cognitive protocol v4, native turn-boundary wake signals on six tools), the supported-tool list is an honest 20 after the Gemini CLI / Continue.dev / Roo Code retirements, and the session bus's exactly-once guarantee was proven by a multi-process chaos harness. This mission checks that all of that holds on a real Linux machine, on your distro, with your shell.

## Steps

```bash
npm install -g @egchq/egc
egc install
egc doctor
cd <a-real-project>
egc init
```

Note: after a bare `egc install`, `egc doctor` reporting "No EGC install-state files found" and a missing `state.db` is expected: the core runtime is installed, nothing more. It is `egc init` (run inside a project) that creates the state store, bootstraps your AI tools and starts the dashboard. That is the moment the two behavior questions below refer to.

If you already have an older version, run the same commands and note anything odd during the transition. If `egc doctor` reports anything unexpected after `egc init`, paste it: that is exactly the kind of thing this mission exists to catch.

## What to report (as a comment here)

- Distro and version, Node.js version, shell, and how Node is installed (distro package, nvm, mise, asdf, ...)
- Output of `egc --version` and the final `egc doctor` summary
- Whether the dashboard opened after `egc init`
- Whether your AI tools picked up EGC afterwards
- Anything that looked wrong, however small

## Why this matters

"Linux" is not one platform. Node installed through nvm, mise, asdf, snap or a distro package puts binaries and global packages in different places, and an installer that assumes one layout silently misbehaves on the others. The maintainer can only test the layout on his own machine; your distro is the one nobody has checked.

No code required. A careful report is the whole contribution, and it gets you credited in the release notes and the Hall of Fame.

rathaur-ankit:
### Environment
- **Distro and version:**  Ubuntu 26.04 LTS
- **Node.js version:**  v22.22.1 (npm 9.2.0)
- **Shell:**  zsh
- **How Node is installed:**  Global/Distro package (installation required `sudo npm install -g`)

### Outputs
**`egc --version`**
```
1.1.18
```

**`egc doctor` summary**
```
No EGC install-state files found for the current home/project context.
[...]

PR #1480, opened 2026-09-19
fix(install): never npm link a package installed under a non-default prefix (#1464)

### LINKED ISSUE #1464: fix(install): egc install links the package into npm root -g when it runs from another prefix
## What happened

Installed the published package under its own prefix and ran the bare install from there:

```bash
npm install -g --prefix ~/egc-e2e/npm @egchq/egc@1.1.22
export PATH=~/egc-e2e/npm/bin:$PATH
egc install
```

macOS 15.7.9 (Intel), Node 20.18.0, npm 10.8.2, default global prefix `/usr/local`. After `egc install`, two symlinks exist that nobody asked for:

```text
/usr/local/lib/node_modules/@egchq/egc -> ../../../../../Users/<user>/egc-e2e/npm/lib/node_modules/@egchq/egc
/usr/local/bin/egc -> ../lib/node_modules/@egchq/egc/scripts/egc.js
```

Everything else passes (`egc doctor`, `egc init --yes` in a fresh project, doctor again). Removing the custom prefix afterwards leaves both links dangling, and `command -v egc` then fails on a machine that never had EGC in `/usr/local`.

Checked step by step: the links do not exist after `npm install -g --prefix`, nor after `egc --version`; they appear during `egc install`, and stay through `egc doctor` and `egc init`. npm alone is not the cause: `npm install -g --prefix /tmp/x cowsay` on the same machine writes nothing to `/usr/local`.

## Where

`scripts/install.sh` lines 176 to 182 and `scripts/install.ps1` lines 222 to 240. Both compute `$(npm root -g)/@egchq/egc` and, when that directory does not resolve to the running tree, run `npm link --silent`. The comment states the intent: a clone should outrank a stale global install left on PATH. A package that is itself an npm install under any other prefix (a one-off `--prefix`, `NPM_CONFIG_PREFIX` set only for the install command, a version manager switched after installing) takes the same branch and gets linked into a global prefix it never lived in.

The same test on `windows-latest` through the published package took the other branch (`egc command already provided by the global npm install`) because the runner uses the default prefix, so Windows is affected only under a non-default prefix.

## Expected

Link only when the running tree is a checkout (has `.git`, or is not inside a `node_modules` directory). An npm-installed package should never call `npm link`, whatever prefix it lives in. A line in the output naming the link when it does happen would also help: today it is silent unless the person reads the full install log.

## Environment

- EGC 1.1.22 from the registry
- macOS 15.7.9, x86_64, Node 20.18.0, npm 10.8.2, prefix `/usr/local`
- Reproduced twice, cleaned by removing the two links by hand

kunaaaalcodes:
PR Description #1464 
## Summary

`egc install` runs `npm link` whenever the running tree is not the default
global npm install. That guard only distinguishes "is already the global
install" from "everything else", and silently assumed everything else is a
from-source checkout.

An npm-installed package under any other prefix — a one-off
`npm install -g --prefix <other>`, `NPM_CONFIG_PREFIX`, or a version manager
switched after installing — takes that same "everything else" branch after
passing the #1218 guard, and gets `npm link`-ed into the machine's default
global prefix. On the reported machine that created two dangling symlinks in
`/usr/local` that nobody asked for, breaking `command -v egc` after the custom
prefix was removed.

This PR adds a real checkout test to both installers so `npm link` only runs
for a git checkout (its tree has a `.git` directory, or a `.git` file in a
worktree). An npm install always lives inside a `node_modules` directory, so
that path alone identifies it regardless of which prefix npm thinks is global.
A successful link now also prints `linked: egc -> <target>` — previously npm
`--silent` made the whole operation invisible unless the install log was read.

## Component Type

- [x] Core Engine / Orchestrator (installer scripts: `scripts/install.sh` and `scripts/install.ps1`)
- [ ] MCP Server (egc-guardian / egc-memory)
- [ ] Agent
- [ ] Skill
- [ ] Hook
- [ ] Command

## Validation

- `scripts/install.sh` and `scripts/install.ps1` both pass syntax checks
  (`bash -n` and the Windows PowerShell parser) on the machines they run on.
- `tests/scripts/install-ps1.test.js`: 17/17 pass, including the new #1464
  parity test that asserts both installers implement the same guard.
- `tests/scripts/install-apply.test.js`: 36/36 and `install-onboarding.test.js`:
  13/13 pass — no regressions in the delegated install path.
- The checkout classifier was functionally exercised against seven path
  shapes: npm install under another prefix (skip), default global prefix
  (skip), `node_modules` as a leaf (skip), real git clone (link), manual copy
  without `.git` (link), git worktree (link), and a nested MSYS path (skip).
- `tests/scripts/install-sh.test.js` self-skips on win32, so it runs for the
  first time on Linux/macOS CI; its source-level assertions for the new guard
  were verified against `scripts/install.sh` directly on Windows.
- Avoided `npm bin` for the success-message path lookup: it was removed in
  npm 11, so the message names the command and target instead.

## Checklist

- [x] All commits carry `Signed-off-by` (DCO, via `.githooks`)
- [ ] `node tests/run-all.js` passes — targeted suites ran green; full suite not executed in this session
- [ ] `npm audit --audit-level=high` — no new dependencies introduced
- [x] Markdownlint — no `.md` files modified
- [x] PR changes fewer than 150 code files (4 files: 2 installers + 2 tests)
- [x] No file shared across concurrent EGC processes touched (no concurrent-access test required)
- [x] Tested on Linux/macOS/Windows — Windows (PowerShell + Git Bash) verified locally; bash installer covered on Linux/macOS by CI
- [x] Doesn't break the `registry/runtime-map.json` registry (untouched)

---
## Summary by cubic
Fixes #1464: `egc install` no longer runs `npm link` on an npm-installed package under a non-default prefix. The installers previously treated any non-default tree as a source checkout, which could create dangling symlinks in the default global prefix and break `command -v egc` after the custom prefix was removed.

**Changes**
- Runs `npm link` only for git checkouts, detected by a `.git` directory or worktree `.git` file.
- Treats any tree inside `node_modules` as an npm install and prints a skip message instead of linking.
- Prints `linked: egc -> <target>` after a successful link.
- Adds parity tests covering the new guard for both `install.sh` and `install.ps1`.

<sup>Written for commit 56a7120901278e711c4521dfabceb5423099d04e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1480?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Fmarzochi:
Thanks, Kunal. I ran the installer for real on this head and on main, with a stub `npm` on PATH that records every `npm link`: from a tree under `<prefix>/lib/node_modules/@egchq/egc`, main links (the #1464 bug) and this branch prints the skip line and never calls link; the default global prefix still takes the #1218 branch; a git worktree (`.git` file), a git clone and a copy without `.git` outside `node_modules` all link and print the new `linked:` line. The logic is right and I have nothing against it.

Two things before I merge, since the installers run on every user's machine:

1. The new comment in both installers says `npm bin` was removed in npm 11. It went away in npm 9.0.0. Please change the two comments to npm 9.

2. The two new tests assert the source text, not the behaviour. I checked by mutation: replacing `elif ! is_checkout; then` with `elif false; then` keeps both tests green, and removing the `node_modules` case from `is_checkout` keeps the install.ps1 parity test green, because `.git` and `node_modules` already appear elsewhere in both scripts. Please add one case that executes `scripts/install.sh` from a path inside `node_modules` with a stub `npm` on PATH and asserts that the skip line is printed and that `npm link` is never called; `tests/scripts/install-sh.test.js` already runs the script through `execFileSync` with a controlled PATH, so the shape is there. And anchor the parity assertion on the real constructs (`[[ -e "$ROOT_DIR/.git" ]]`, the `*/node_modules` case, `Test-Path (Join-Path $RootDir ".git")`, the `[\\/]node_modules` match) instead of the bare substrings.

On the other bot comments: the two CodeRabbit "Major" items are not for this PR. A copy without `.git` outside `node_modules` is meant to link, that is the expected behaviour written in #1464 and what the installer did before this change; and there is no npm binary-resolution pattern in these scripts to follow, every npm call in them is plain.

Fmarzochi:
Kunal, one thing outside this PR: the e-mail I sent you on the 19th with the thirteen decisions is stuck. Gmail reports your inbox is out of storage space (452 4.2.2) and will keep retrying for two more days before giving up. If you free some space, or tell me another address here, it goes through. Nothing changes for this PR: the two adjustments above are still the only thing between it and the merge.

Issue #1460, opened 2026-09-16
test(mcp-health-check): the stderr-probe case gives the probe 100 ms and a cold macos-latest runner declares the failing server healthy before it exits (main at ffc79f1b)


Fmarzochi:
## What happened

On the CI run for main at ffc79f1b (the squash of #1456, a lockfile-only bump in `mcp/servers/egc-guardian`), the job `Test (macos-latest, Node 22.x, yarn)` failed in `tests/hooks/mcp-health-check.test.js`:

```
✗ includes command stderr and config env in unhealthy probe reasons
    Error: Expected stderr probe failure to block
```

4158 of 4159 tests passed. The same job on the PR head had passed minutes earlier, the two commits merged right after it (c644f777 and 33a2786d) passed the same lane on the same test tree, and #1456 does not touch this test or the hook. A rerun of the failed job (attempt 2 of run 35124106760) passed, so main is green again and this was cold-runner timing, not the code. The issue stays open for the test-side fix described below.

## Reading

The case starts a stdio server that prints to stderr and exits 1 at once, sets `ECC_MCP_HEALTH_TIMEOUT_MS=100`, and expects the hook to block (exit code 2) with the stderr text in the reason. In `scripts/hooks/mcp-health-check.js`, `probeCommandServer` waits `timeoutMs` and, if the child has not exited yet, arms a grace timer of another `timeoutMs`; a child still alive at the end of that grace window is declared healthy ("accepted a new stdio process") and the hook exits 0. On a cold macos-latest runner, spawning Node for the stub server can take longer than the 200 ms the two windows add up to, so the probe declares the failing server healthy before it manages to fail, and the assertion sees exit code 0 instead of 2.

The sibling cases in the same file that assert an unhealthy verdict run on 1000 ms; this is the only unhealthy assertion left on the 100 ms budget. It is the fixed-budget-sized-on-Linux pattern of #1448 and of the fixes in #1203, #1209 and #1212, again in a test rather than in product code.

## Suggested fix

Raise the budget of this case to the 1000 ms its siblings use (1500 ms if you want more margin). The server under test exits immediately, so the case stays as fast as it is today; only the ceiling moves. The three cases that assert a healthy server on 100 ms can stay as they are, since a short window can only make a healthy verdict arrive sooner.

Optional second step: make the assertion print `result.code` and `result.stderr` when it fails, so the next occurrence names what the hook actually decided instead of a bare "expected to block".

[...]

Issue #1448, opened 2026-09-14
test(opencode-plugin): the prompt-timeout test shares its 150 ms budget with the bridge spawn and fails for the wrong reason on a cold windows-latest bun runner (main at 66cc67da)


Fmarzochi:
## What happened

On the CI run for main at 66cc67da (the squash of #1447), the job `Test (windows-latest, Node 22.x, bun)` failed in `tests/hooks/opencode-egc-plugin.test.js`:

```
✗ fails open when the OpenCode prompt exceeds its timeout
    Error: Expected values to be strictly equal:
    0 !== 1
```

The same job on the PR head d870c77f, an identical tree, had passed minutes earlier, and #1447 only touches `tests/scripts/install-ps1.test.js`. A rerun of the failed job is in progress; this issue records the mechanism regardless of its outcome.

## Reading

The test sets `EGC_SESSION_CONTEXT_TIMEOUT_MS=150` and expects the plugin to reach `client.session.prompt` (a promise that never resolves) and then fail open on the prompt timeout, asserting `attempts === 1`. `attempts` was 0: the plugin failed open one stage earlier. The plugin reads that single variable for both budgets, the bridge subprocess (`scripts/hooks/opencode-egc-plugin.js`, the `withTimeout` around the bridge run) and the prompt, so on a runner where spawning the Node stub bridge alone takes more than 150 ms (a cold Windows runner under bun), the bridge times out first, the context is empty, and the prompt is never attempted. The test then fails for the wrong reason. This is the fixed-budget-sized-on-Linux pattern already fixed in #1203, #1209 and #1212, here in a test rather than in product code.

## Suggested fix

Give the bridge stage room and keep only the prompt on the short leash. Two shapes work:

- in the test, raise `EGC_SESSION_CONTEXT_TIMEOUT_MS` to something a cold spawn cannot exceed (1500 ms) and widen the upper bound of the elapsed assertion accordingly (5000 ms), so the bridge completes and the prompt is the stage that times out; the sibling test "fails open when the bridge exceeds its timeout" keeps its short budget because there the bridge is the stage under test;
- or let the plugin take a separate prompt budget (a second environment variable with the same default), so each stage can be exercised on its own.

The first is a two-line test change; the second is a small product change and belongs in its own conversation. Either way, the assertion should name which stage timed out when it fails, so the next occurrence is not another `0 !== 1`.

Run: https://github.com/Fmarzochi/EGC/actions/runs/34798340684

[...]

PR #1481, opened 2026-09-19
test(mcp-health-check): give the stderr-probe case a 1000 ms budget like its siblings (#1460)

kunaaaalcodes:
## Summary #1460 

`tests/hooks/mcp-health-check.test.js` has one test — "includes command stderr
and config env in unhealthy probe reasons" — that asserts an **unhealthy**
verdict on the 100 ms probe budget its healthy siblings use. That budget has a
hidden multiplier: `probeCommandServer` waits `timeoutMs`, then arms a grace
window of *another* `timeoutMs` before declaring a still-running child healthy.
At 100 ms the failing stub server has only 200 ms total to spawn Node and exit.
On a cold macos-latest runner the spawn can take longer than that, so the probe
declared the failing server healthy ("accepted a new stdio process") and the
assertion saw exit code 0 instead of 2 (CI run 35124106760, main @ ffc79f1b).
A rerun passed, confirming it was cold-runner timing and not the hook.

This is the fixed-budget pattern of #1448 recurring in a test rather than in
product code, so the fix is test-side only:

- Raise this case to the **1000 ms budget its sibling unhealthy cases already
  use** (lines 446/482/514/523/634/697). The server under test exits
  immediately, so the case's wall time is unchanged — only the ceiling moves.
- Make the assertion print `result.code` and `result.stderr` when it fails, so
  the next occurrence names what the hook actually decided instead of a bare
  "Expected stderr probe failure to block".

The three cases that assert a *healthy* server on 100 ms (lines 249/383/413)
are intentionally left alone: a short window can only make a healthy verdict
arrive sooner, never flip it. A grep confirms no unhealthy assertion remains on
the short budget.

## Component Type

Test-only change to `tests/hooks/mcp-health-check.test.js`; product code
(`scripts/hooks/mcp-health-check.js`) is untouched.

## Validation

- `node tests/hooks/mcp-health-check.test.js` → `Results: Passed: 20, Failed: 0`.
- `npx eslint tests/hooks/mcp-health-check.test.js` → no new problems (the
  single complexity warning on `runTests` is pre-existing; this change adds no
  control flow).
- Grep audit: the only remaining 100 ms cases are the three healthy-assertion
  cases; all unhealthy assertions now run on 1000/2000 ms.
- The hook's two-window logic was read and confirmed against
  `scripts/hooks/mcp-health-check.js` (lines 333–388): a child exiting in
  either window already resolves unhealthy via the `exit` listener; the flake
  was solely the 200 ms total ceiling on a cold runner.

## Checklist

- [x] All commits carry `Signed-off-by` (DCO, via `.githooks`)
- [ ] `node tests/run-all.js` passes — targeted suite green (20/20); full suite not executed in this session
- [x] `npm audit` — no dependencies introduced or changed
- [x] Markdownlint — no `.md` files modified
- [x] PR changes fewer than 150 code files (1 file)
- [x] No file shared across concurrent EGC processes touched (no concurrent-access test required)
- [x] Tested on Linux/macOS/Windows — runs green on the local Windows runner; the raised ceiling directly de-risks the failing macos-latest lane
- [x] Doesn't break the `registry/runtime-map.json` registry (untouched)

---
## Summary by cubic
Fixes the flaky stderr-probe test in `mcp-health-check` by raising its timeout budget from 100 ms to 1000 ms, matching the sibling unhealthy cases, and makes CI retry `corepack prepare yarn@stable --activate` up to three times. `probeCommandServer` waits `timeoutMs` plus another `timeoutMs` grace window, so 100 ms gave only 200 ms to spawn Node and exit — too tight on a cold macos runner. The assertion now prints the actual exit code and stderr on failure. Resolving 'stable' polls `repo.yarnpkg.com/tags`, a fetch that occasionally fails on cold runners.

<sup>Written for commit 83db81ca9e2b22d3b86dd95bf9f6db7e1944db67. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Fmarzochi/EGC/pull/1481?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Fmarzochi (approved):
Reviewed with the hook run for real: a stdio server that takes 250 ms to fail is declared healthy on the 100 ms budget (exit 0) and blocked on 1000 ms (exit 2), exactly the window the issue describes, and the suite passes 20 of 20 on this head. The corepack retry answers the Internal Error your first push hit on the Windows yarn lane, and the step already runs under bash on every OS, so it stays. One cosmetic point for a later pass: the loop still sleeps five seconds after the third attempt. Thanks, Kunal, merging. The changelog and roadmap entries go in a separate docs PR on my side, as usual.

